PR #1, opened 2019-04-30
Update script to support better options and ignore line numbers for diff

danebou:
- Expose more options via command line args
- Fix bug where line numbers were being included in the register diff

### DIFF
--- a/diff.sh
+++ b/diff.sh
@@ -86,7 +86,7 @@ if [[ $DIFF_OBJ = 1 ]]; then
     OBJDUMP="mips-linux-gnu-objdump -drz"
     $OBJDUMP $REFOBJFILE | grep "<$1>:" -A1000 > $BASEDUMP
     $OBJDUMP $OBJFILE | grep "<$1>:" -A1000 > $MYDUMP
-    DIFF_ARGS+=" -o --stop-jr-ra"
+    DIFF_ARGS+=" -o"
 else
     if [[ $MAKE = 1 ]]; then
         make $MAKEFLAGS "$MYIMG"
--- a/diff.sh
+++ b/diff.sh
@@ -7,6 +7,7 @@ set -e
 DIFF_OBJ=0
 IGNORE_REGS=0
 MAKE=0
+DIFF_ARGS="-l"
 
 POSITIONAL=()
 while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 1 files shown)


diff.sh
```diff
@@ -86,7 +86,7 @@ if [[ $DIFF_OBJ = 1 ]]; then
     OBJDUMP="mips-linux-gnu-objdump -drz"
     $OBJDUMP $REFOBJFILE | grep "<$1>:" -A1000 > $BASEDUMP
     $OBJDUMP $OBJFILE | grep "<$1>:" -A1000 > $MYDUMP
-    DIFF_ARGS+=--diff-obj
+    DIFF_ARGS+=" -o --stop-jr-ra"
```
simonlindholm: I guess we should really have bash flags for this, or pass unused bash flags on to the diff script, but ... eh, I'm fine with being lazy.

diff.sh
```diff
@@ -118,12 +118,10 @@ import re
 import string
 from signal import signal, SIGPIPE, SIG_DFL
 
-leftright_spacing = 40
-
 # Fixes pipe error
-signal(SIGPIPE,SIG_DFL)
+signal(SIGPIPE,SIG_DFL)
```
simonlindholm: trailing whitespace!!!1!

diff.sh
```diff
@@ -245,15 +242,18 @@ def process(lines, options):
         # Replace tabs with spaces
         diff_rows.append(row)
         originals.append(original)
+        line_nums.append(line_num)
+        if options.stop_jrra and mnemonic == 'jr' and row.split('\t')[1].strip() == 'ra':
```
simonlindholm: Functions could contain more than one `jr ra`... but I guess when that happens it's easy to turn this behavior off, and such functions are typically small since they don't contain any register cleanup, and small functions are less often diffed. The alternative would be to have the `--stop-jr-ra` flag take an optional count parameter, but again, I'm perfectly fine with being lazy here and keeping this as is.

danebou: Fair. I'll leave this off by default then. That way its there if we need it but avoids cutting off the function.

diff.sh
danebou: What do you mean bash flags? But i definitely agree with passing the unused bash flags.

simonlindholm: I having something like
```
-s)
  DIFF_ARGS+=" --stop-jr-ra"
  shift
```
in the switch at the top of the script. Passing on unused flags is slightly more complex because diff-settings.sh may also parse flags, so you need two switches, but on the whole it's not so bad.

PR #19, opened 2020-06-08
Add new "diff ELF" mode for diffing stripped and non-stripped ELF

leoetlino:
This is useful for diffing functions between two ELFs, in the situation
where the first one is stripped and the other one is non-stripped,
using start addresses to locate functions for the former and symbols
for the latter.

It's somewhat of a mix of the two existing modes, but I think making
a new mode is cleaner.

(My personal use case for this is diffing BotW using the original
executable and a ELF containing only the decompiled functions.
For various reasons, disassembling everything like the OoT decomp
project is annoying and inconvenient.)

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -50,6 +50,11 @@ parser.add_argument(
     help="Diff .o files rather than a whole binary. This makes it possible to see symbol names. (Recommended)",
 )
 parser.add_argument(
+    "-e",
+    dest="diff_elf_symbol",
+    help="Diff a given function in two ELFs, one being stripped and the other one non-stripped. Requires objdump from binutils 2.33+.",
+)
+parser.add_argument(
     "--base-asm",
     dest="base_asm",
     metavar="FILE",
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -291,6 +296,32 @@ def search_map_file(fn_name):
...
+def dump_elf():
+    if not baseimg or not myimg:
+        fail("Missing myimg/baseimg in config.")
+    if base_shift:
+        fail("--base-shift not compatible with -e")
+    if args.end is None:
+        fail("The end address is required.")
+
+    start_addr = eval_int(args.start)
+    end_addr = eval_int(args.end)
```
simonlindholm, line 308: these should take error messages, otherwise eval_int returns None if it fails to parse

simonlindholm: could be worth doing
```py
    if args.end is not None:
        end_addr = eval_int(args.end, "End address must be an integer expression.")
    else:
        end_addr = start_addr + MAX_FUNCTION_SIZE_BYTES
```
just for consistency

diff.py
```diff
@@ -49,6 +49,11 @@ def fail(msg):
     action="store_true",
     help="Diff .o files rather than a whole binary. This makes it possible to see symbol names. (Recommended)",
 )
+parser.add_argument(
+    "-e",
+    dest="diff_elf_symbol",
+    help="Diff a function in two ELFs, one being stripped and the other one non-stripped. Requires objdump from binutils 2.33+.",
```
simonlindholm, line 55: Extend this help text to explain that a symbol name should be given

PR #20, opened 2020-06-08
Max lines

onlymx13:
I added an argument for the maximum number of lines. I guess that's it. This PR is black-formatted.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -137,7 +137,7 @@ parser.add_argument(
     dest="max_lines",
     type=int,
     default=1024,
-    help="The maximum length of the diff, in lines. Not recommended when -f is used.",
+    help="The maximum length of the diff, in lines.",
 )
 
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -131,6 +131,15 @@ def fail(msg):
     help="Diff algorithm to use.",
 )
 
+parser.add_argument(
+    "--max-size",
+    "--max-lines",
+    dest="max_lines",
+    type=int,
+    default=1024,
+    help="The maximum length of the diff, in lines. Not recommended when -f is used.",
```
simonlindholm, line 140:
```suggestion
    help="The maximum length of the diff, in lines. Not recommended when -s is used.",
```

(or drop the note about -s altogether; currently it *is* actually useful to combine them, since the -s filter is applied after the disassembly limit)

PR #22, opened 2020-06-08
Add an option to show source code (if possible)

leoetlino:
This adds a ``--source`` option to show source code intermixed with assembly
whenever possible, which usually helps identify where each part
of the function comes from. This is especially useful for C++ code,
which often makes heavy use of inlined functions, templates and
often looks very different from the original source code.

Example (featuring all of the above):

image

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -5,6 +5,7 @@ import os
 import ast
 import argparse
 import subprocess
+import collections
 import difflib
 import string
 import itertools
@@ -53,6 +54,11 @@ parser.add_argument(
     "-e",
     dest="diff_elf_symbol",
     help="Diff a given function in two ELFs, one being stripped and the other one non-stripped. Requires objdump from binutils 2.33+.",
+)
+parser.add_argument(
+    "--source",
+    action="store_true",
+    help="Show source code (if possible). Only works with -o and -e.",
 )
 parser.add_argument(
     "--base-asm",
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -5,12 +5,14 @@
...
 import argparse
 import subprocess
+import collections
 import difflib
 import string
 import itertools
 import threading
 import queue
 import time
+import cxxfilt
```
simonlindholm, line 15: as with `import Levenshtein` it might be nice to make this dependency optional to spare people the pain of installing new packages, and improving startup time a little

diff.py
```diff
@@ -831,9 +856,28 @@ def do_diff(basedump, mydump):
                         branch_targets2[j1 + k] + ":", "~>"
                     )
 
+            if args.source and comments2[j1 + k] is not None:
+                out2 += f" {comments2[j1 + k][0]}"
```
simonlindholm, line 860: only do this if `has2`, otherwise this may produce incorrect results or crash

Are these comments useful? The image just shows unmangled function names

simonlindholm:
I should add, this looks fantastic.

diff.py
leoetlino: Yeah, they're useful for function calls:

      2738:    bl      216c     <memcpy@plt> ~>       

Sometimes the function call appears in the source code, but not always:

image

when the call is implicitly generated (when calling destructors of parent classes, or calling constructors for member variables for example).

simonlindholm: Ah, I see. For `bl` it might be nicer to parse out the comment similar to a relocation and present it as part of the asm (not just with --source), but for tail recursion that does get a bit trickier... That adrp comment looks vaguely useful as well

PR #23, opened 2020-06-08
Ignore immediate diffs for branches if they have identical targets

leoetlino:
objdump prints branch targets using absolute addresses which means that
whenever the reference function and the decomp function do not have the
same base address, branches that are actually identical (with the
same PC-relative offset) will still be detected as being different.

This can be fixed by deriving the relative offset from the line number
and the absolute address that is given to us by objdump, and only
comparing the relative offsets instead.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -237,6 +237,10 @@ def eval_int(expr, emsg=None):
         return None
 
 
+def eval_line_num(expr):
+    return int(expr.strip().replace(":", ""), 16)
+
+
 def run_make(target, capture_output=False):
     if capture_output:
         return subprocess.run(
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -782,13 +786,23 @@ def do_diff(basedump, mydump):
...
-                    branch1, branch2 = color_branch_imms(branch1, branch2)
+
+                    same_relative_target = False
+                    if branch_targets1[i1 + k] is not None and branch_targets2[j1 + k] is not None:
+                        relative_target1 = eval_line_num(branch_targets1[i1 + k]) - eval_line_num(line_num1)
+                        relative_target2 = eval_line_num(branch_targets2[j1 + k]) - eval_line_num(line_num2)
+                        same_relative_target = relative_target1 == relative_target2
+
+                    if not same_relative_target:
+                        branch1, branch2 = color_branch_imms(branch1, branch2)
```
simonlindholm, line 797: I feel like the check for branch1 == branch2 within color_branch_imms is a bit sketchy with this change... OTOH, it probably doesn't hurt in practice (might even help, in the non-`-e` case), and the `maybe_normalize_large_imms(original1) == maybe_normalize_large_imms(original2)` check is equally broken. So eh, let's go with this I guess.

PR #27, opened 2020-07-22
Fix duplication of source lines when "!has2"

leoetlino:
Only print source lines if "has2", otherwise the same source lines can
end up being printed several times when some lines are missing
on the right-hand side.


simonlindholm:
good catch

PR #31, opened 2020-08-30
tab-completion

Dragorn421:
The first commit actually adds tab-completion, the second does reordering for performance.

Soft requirement: `pip install argcomplete` (does nothing different compared to current master if argcomplete isn't available)
See [the first commit](https://github.com/simonlindholm/asm-differ/commit/88e64aa81e8d3a1f37d73d4e33af0b9ba219a5ca)

How the script is called matters, if tab completion has been setup for `./diff.py` using something else like `../asm-differ/diff.py` won't allow tab-completion. I think `argcomplete` intends to have the scripts in the path.

Symbol detection is very rough (a somewhat equivalent regex would be `(?: |^)([^ \n]+)(?: |$)` maybe?), but it seems ok

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1,39 +1,9 @@
 #!/usr/bin/env python3
 import sys
-import re
-import os
-import ast
-try:
-    import argcomplete
-except ModuleNotFoundError:
-    argcomplete = None
-import argparse
-import subprocess
-import difflib
-import string
-import itertools
-import threading
-import queue
-import time
-from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, Union
-
 
 def fail(msg):
     print(msg, file=sys.stderr)
     sys.exit(1)
-
-
-MISSING_PREREQUISITES = (
-    "Missing prerequisite python module {}. "
-    "Run `python3 -m pip install --user colorama ansiwrap watchdog python-Levenshtein cxxfilt` to install prerequisites (cxxfilt only needed with --source)."
-)
-
-try:
-    from colorama import Fore, Style, Back  # type: ignore
-    import ansiwrap  # type: ignore
-    import watchdog  # type: ignore
-except ModuleNotFoundError as e:
-    fail(MISSING_PREREQUISITES.format(e.name))
 
 # Prefer to use diff_settings.py from the current working directory
 sys.path.insert(0, ".")
@@ -41,8 +11,15 @@ try:
     import diff_settings
 except ModuleNotFoundError:
     fail("Unable to find diff_settings.py in the same directory.")
+sys.path.pop(0)
 
-# ==== CONFIG ====
+# ==== COMMAND-LINE ====
+
+try:
+    import argcomplete
+except ModuleNotFoundError:
+    argcomplete = None
+import argparse
 
 parser = argparse.ArgumentParser(description="Diff MIPS assembly.")
 
@@ -201,6 +178,35 @@ if hasattr(diff_settings, "add_custom_arguments"):
 
 if argcomplete:
     argcomplete.autocomplete(parser)
+
+# ==== IMPORTS ====
+
+import re
+import os
+import ast
+import subprocess
+import difflib
+import string
+import itertools
+import threading
+import queue
+import time
+from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, Union
+
+
+MISSING_PREREQUISITES = (
+    "Missing prerequisite python module {}. "
+    "Run `python3 -m pip install --user colorama ansiwrap watchdog python-Levenshtein cxxfilt` to install prerequisites (cxxfilt only needed with --source)."
+)
+
+try:
+    from colorama import Fore, Style, Back  # type: ignore
+    import ansiwrap  # type: ignore
+    import watchdog  # type: ignore
+except ModuleNotFoundError as e:
+    fail(MISSING_PREREQUISITES.format(e.name))
+
+# ==== CONFIG ====
 
 args = parser.parse_args()
 
--- a/diff.py
+++ b/diff.py
@@ -27,6 +27,10 @@ start_argument = parser.add_argument("start", help="Function name or address to 
 if argcomplete:
     def complete_symbol(**kwargs):
         prefix = kwargs["prefix"]
+        if prefix == "":
+            # skip reading the map file, which would
+            # result in a lot of useless completions
+            return []
         parsed_args = kwargs["parsed_args"]
         config = {}
         diff_settings.apply(config, parsed_args)
--- a/diff.py
+++ b/diff.py
@@ -36,22 +36,22 @@ if argcomplete:
         completes = []
         with open(mapfile) as f:
             data = f.read()
-            if data.startswith(prefix):
-                endPos = data.find(" ")
-                completes.append(data if endPos == -1 else data[:endPos])
+            # assume symbols are prefixed by a space character
             search = f" {prefix}"
             pos = data.find(search)
             while pos != -1:
                 # skip the space character in the search string
                 pos += 1
-                endPos = data.find(" ", pos)
+                # assume symbols are suffixed by either a space
+                # character or a (unix-style) line return
+                endPos = min(data.find(" ", pos), data.find("\n", pos))
                 if endPos == -1:
                     match = data[pos:]
                     pos = -1
                 else:
                     match = data[pos:endPos]
                     pos = data.find(search, endPos)
-                completes.append(match.strip())
+                completes.append(match)
         return completes
     start_argument.completer = complete_symbol
 
--- a/diff.py
+++ b/diff.py
@@ -48,7 +48,14 @@ if argcomplete:
                 pos += 1
                 # assume symbols are suffixed by either a space
                 # character or a (unix-style) line return
-                endPos = min(data.find(" ", pos), data.find("\n", pos))
+                spacePos = data.find(" ", pos)
+                lineReturnPos = data.find("\n", pos)
+                if lineReturnPos == -1:
+                    endPos = spacePos
+                elif spacePos == -1:
+                    endPos = lineReturnPos
+                else:
+                    endPos = min(spacePos, lineReturnPos)
                 if endPos == -1:
                     match = data[pos:]
                     pos = -1
--- a/diff.py
+++ b/diff.py
@@ -3,6 +3,10 @@ import sys
 import re
 import os
 import ast
+try:
+    import argcomplete
+except ModuleNotFoundError:
+    argcomplete = None
 import argparse
 import subprocess
 import difflib
@@ -41,7 +45,35 @@ except ModuleNotFoundError:
 # ==== CONFIG ====
 
 parser = argparse.ArgumentParser(description="Diff MIPS assembly.")
-parser.add_argument("start", help="Function name or address to start diffing from.")
+
+start_argument = parser.add_argument("start", help="Function name or address to start diffing from.")
+if argcomplete:
+    def complete_symbol(**kwargs):
+        prefix = kwargs['prefix']
+        parsed_args = kwargs['parsed_args']
+        completes = []
+        config = dict()
+        diff_settings.apply(config, parsed_args)
+        mapfile = config.get('mapfile')
+        if not mapfile:
+            return []
+        with open(mapfile) as f:
+            for line in f:
+                pos = line.find(prefix)
+                # if found the prefix at start of a word
+                # (start of line or if there's a space before)
+                if pos >= 0 and (pos == 0 or line[pos-1] == ' '):
+                    symbolEndPos = line.find(' ', pos)
+                    if symbolEndPos == -1:
+                        # -1 strips the line return '\n' (assuming Unix)
+                        symbol = line[pos:-1]
+                    else:
+                        symbol = line[pos:symbolEndPos]
+                    completes.append(symbol)
+        return completes
+    start_argument.completer = complete_symbol
+del start_argument
+
 parser.add_argument("end", nargs="?", help="Address to end diff at.")
 parser.add_argument(
     "-o",
@@ -166,6 +198,9 @@ parser.add_argument(
 # Project-specific flags, e.g. different versions/make arguments.
 if hasattr(diff_settings, "add_custom_arguments"):
     diff_settings.add_custom_arguments(parser)  # type: ignore
+
+if argcomplete:
+    argcomplete.autocomplete(parser)
 
 args = parser.parse_args()
 
--- a/diff.py
+++ b/diff.py
@@ -16,7 +16,7 @@ sys.path.pop(0)
 # ==== COMMAND-LINE ====
 
 try:
-    import argcomplete
+    import argcomplete  # type: ignore
 except ModuleNotFoundError:
     argcomplete = None
 import argparse
@@ -26,30 +26,34 @@ parser = argparse.ArgumentParser(description="Diff MIPS assembly.")
 start_argument = parser.add_argument("start", help="Function name or address to start diffing from.")
 if argcomplete:
     def complete_symbol(**kwargs):
-        prefix = kwargs['prefix']
-        parsed_args = kwargs['parsed_args']
-        completes = []
-        config = dict()
+        prefix = kwargs["prefix"]
+        parsed_args = kwargs["parsed_args"]
+        config = {}
         diff_settings.apply(config, parsed_args)
-        mapfile = config.get('mapfile')
+        mapfile = config.get("mapfile")
         if not mapfile:
             return []
+        completes = []
         with open(mapfile) as f:
-            for line in f:
-                pos = line.find(prefix)
-                # if found the prefix at start of a word
-                # (start of line or if there's a space before)
-                if pos >= 0 and (pos == 0 or line[pos-1] == ' '):
-                    symbolEndPos = line.find(' ', pos)
-                    if symbolEndPos == -1:
-                        # -1 strips the line return '\n' (assuming Unix)
-                        symbol = line[pos:-1]
-                    else:
-                        symbol = line[pos:symbolEndPos]
-                    completes.append(symbol)
+            data = f.read()
+            if data.startswith(prefix):
+                endPos = data.find(" ")
+                completes.append(data if endPos == -1 else data[:endPos])
+            search = f" {prefix}"
+            pos = data.find(search)
+            while pos != -1:
+                # skip the space character in the search string
+                pos += 1
+                endPos = data.find(" ", pos)
+                if endPos == -1:
+                    match = data[pos:]
+                    pos = -1
+                else:
+                    match = data[pos:endPos]
+                    pos = data.find(search, endPos)
+                completes.append(match.strip())
         return completes
     start_argument.completer = complete_symbol
-del start_argument
 
 parser.add_argument("end", nargs="?", help="Address to end diff at.")
 parser.add_argument(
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -1,47 +1,56 @@
...
+                    symbolEndPos = line.find(' ', pos)
+                    if symbolEndPos == -1:
+                        # -1 strips the line return '\n' (assuming Unix)
+                        symbol = line[pos:-1]
+                    else:
+                        symbol = line[pos:symbolEndPos]
+                    completes.append(symbol)
+        return completes
+    start_argument.completer = complete_symbol
+del start_argument
```
simonlindholm, line 52:
```suggestion
```

diff.py
```diff
@@ -1,47 +1,56 @@
...
+                # if found the prefix at start of a word
+                # (start of line or if there's a space before)
+                if pos >= 0 and (pos == 0 or line[pos-1] == ' '):
+                    symbolEndPos = line.find(' ', pos)
+                    if symbolEndPos == -1:
+                        # -1 strips the line return '\n' (assuming Unix)
+                        symbol = line[pos:-1]
+                    else:
+                        symbol = line[pos:symbolEndPos]
+                    completes.append(symbol)
```
simonlindholm, line 49:
```suggestion
                    completes.append(line[pos:].split()[0] or "")
```

Alternatively, how about using a real regex instead, combined with `re.escape()`? Seems like it might be a bit faster.

diff.py
```diff
@@ -1,47 +1,56 @@
...
+    argcomplete = None
+import argparse
 
 parser = argparse.ArgumentParser(description="Diff MIPS assembly.")
-parser.add_argument("start", help="Function name or address to start diffing from.")
+
+start_argument = parser.add_argument("start", help="Function name or address to start diffing from.")
+if argcomplete:
+    def complete_symbol(**kwargs):
+        prefix = kwargs['prefix']
```
simonlindholm, line 29:
```suggestion
        prefix = kwargs["prefix"]
```
(keeping with [black](https://github.com/psf/black)'s standard of only using " quotes)

diff.py
```diff
@@ -1,47 +1,56 @@
...
+        parsed_args = kwargs['parsed_args']
+        completes = []
+        config = dict()
```
simonlindholm, line 32:
```suggestion
        config = {}
```

diff.py
```diff
@@ -1,47 +1,56 @@
...
     import diff_settings
 except ModuleNotFoundError:
     fail("Unable to find diff_settings.py in the same directory.")
+sys.path.pop(0)
 
-# ==== CONFIG ====
+# ==== COMMAND-LINE ====
+
+try:
+    import argcomplete
```
simonlindholm, line 19:
```suggestion
    import argcomplete  # type: ignore
```
for mypy's sake

simonlindholm (commented):
Really nice! Can you add some words about this to the readme as well?

diff.py
Dragorn421: what is mypy?

simonlindholm: https://github.com/python/mypy/

Python type-checker; I use it for most projects. (Just started introducing type annotations into diff.py, so there aren't very many yet.)

diff.py
Dragorn421: <details>

<summary>symbolcompletespeedtest.py</summary>

```py
import time
import re

warmUpUntil = time.time() + 1
while time.time() < warmUpUntil:
    time.sleep(0.001)

n = 20
prefix = 'EnBox'

start = time.time()

for i in range(n):
    completes = []
    with open('oot/build/z64.map') as f:
        for line in f:
            pos = line.find(prefix)
            # if found the prefix at start of a word
            # (start of line or if there's a space before)
            if pos >= 0 and (pos == 0 or line[pos-1] == " "):
                symbolEndPos = line.find(" ", pos)
                if symbolEndPos == -1:
                    # -1 strips the line return '\n' (assuming Unix)
                    symbol = line[pos:-1]
                else:
                    symbol = line[pos:symbolEndPos]
                completes.append(symbol)

end = time.time()

print('find/slice > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open('oot/build/z64.map') as f:
        for line in f:
            pos = line.find(prefix)
            # if found the prefix at start of a word
            # (start of line or if there's a space before)
            if pos >= 0 and (pos == 0 or line[pos-1] == " "):
                completes.append(line[pos:].split()[0] or "")

end = time.time()

print('find/split > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open('oot/build/z64.map') as f:
        for line in f:
            pass

end = time.time()

print('read file > time per read:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r'(?: |^)([^ ]+)(?: |\n)')
    completes = []
    with open('oot/build/z64.map') as f:
        for line in f:
            symbols = prog.findall(line)
            completes.extend(symbol for symbol in symbols if symbol.startswith(prefix))

end = time.time()

print('read file > time per read:', (end - start) / n * 1000, 'ms')
```

</details>

```
$ python3 symbolcompletespeedtest.py
find/slice > time per search: 198.33180904388428 ms
find/split > time per search: 198.94893169403076 ms
read file > time per read: 90.79512357711792 ms
read file > time per read: 1650.8959531784058 ms
```

I would go with `line[pos:].split()[0] or ""` as you suggested then, but what is the purpose of `or ""`?
How would you use regexes efficiently?

simonlindholm: > what is the purpose of or ""?

oh I was just trying to maintain the exact behavior. If `prefix` consists only of whitespace, it could end up with None there. This probably doesn't matter at all and could be handled in better ways than `or ""`...

> How would you use regexes efficiently?

I was thinking something like
```py
data = f.read()
for m in re.finditer(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)", data, re.MULTILINE):
    ... m.group(1)
```

Using just a single call into the regex engine and doing O(result size) work in Python should make this faster than your benchmarked version.

Dragorn421: <details>
<summary>symbolcompletespeedtest.py (click to expand)</summary>

```py
import time
import re

warmUpUntil = time.time() + 1
while time.time() < warmUpUntil:
    time.sleep(0.001)

n = 20
mapFile = 'z64.map'
prefix = 'EnBox'

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        for line in f:
            pos = line.find(prefix)
            # if found the prefix at start of a word
            # (start of line or if there's a space before)
            if pos >= 0 and (pos == 0 or line[pos-1] == " "):
                symbolEndPos = line.find(" ", pos)
                if symbolEndPos == -1:
                    # -1 strips the line return '\n' (assuming Unix)
                    symbol = line[pos:-1]
                else:
                    symbol = line[pos:symbolEndPos]
                completes.append(symbol)

end = time.time()

print('find/slice > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        for line in f:
            pos = line.find(prefix)
            # if found the prefix at start of a word
            # (start of line or if there's a space before)
            if pos >= 0 and (pos == 0 or line[pos-1] == " "):
                completes.append(line[pos:].split()[0] or "")

end = time.time()

print('find/split > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        for line in f:
            pass

end = time.time()

print('read file > time per read:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r'(?: |^)([^ ]+)(?: |\n)')
    completes = []
    with open(mapFile) as f:
        for line in f:
            symbols = prog.findall(line)
            completes.extend(symbol for symbol in symbols if symbol.startswith(prefix))

end = time.time()

print('regex findall symbols then filter > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        data = f.read()
        completes = [m.group(1) for m in re.finditer(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)", data, re.MULTILINE)]

end = time.time()

print('regex finditer prefix-in-regex > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)")
    completes = []
    with open(mapFile) as f:
        data = f.read()
        completes.extend(m.group(1) for m in prog.finditer(data, re.MULTILINE))

end = time.time()

print('regex finditer prefix-in-regex compile, extend (not much different) > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)")
    completes = []
    with open(mapFile) as f:
        for l in f:
            completes.extend(m.group(1) for m in prog.finditer(l))

end = time.time()

print('regex finditer prefix-in-regex per-line > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        data = f.read()
        pos = data.find(prefix)
        # handle the case where the symbol can be found
        # right at the start of the file separately
        if pos == 0:
            endPos = data.find(" ")
            if endPos == -1:
                completes.append(data)
                pos = -1
            else:
                completes.append(data[:endPos])
                pos = data.find(prefix, endPos)
        # pos != 0
        prefix = f" {prefix}"
        while pos != -1:
            pos += 1
            endPos = line.find(" ", pos)
            if endPos == -1:
                completes.append(line[pos:])
                pos = -1
            else:
                completes.append(line[pos:endPos])
                pos = line.find(prefix, endPos)

end = time.time()

print('find/slice f.read() > time per search:', (end - start) / n * 1000, 'ms')
```

</details>

```
# same tests as previously
find/slice > time per search: 368.3366537094116 ms
find/split > time per search: 362.1750473976135 ms
read file > time per read: 321.9892382621765 ms
regex findall symbols then filter > time per search: 996.9877362251282 ms
# finditer
regex finditer prefix-in-regex > time per search: 461.04514598846436 ms
regex finditer prefix-in-regex compile, extend (not much different) > time per search: 500.27135610580444 ms
regex finditer prefix-in-regex per-line > time per search: 1025.657308101654 ms
# f.read() + find() + slicing
find/slice f.read() > time per search: 68.25450658798218 ms
```

I'll definitely go with that last attempt unless you can come up with actually fast regex usage, but would you have ideas on making it look cleaner:
```py
data = f.read()
pos = data.find(prefix)
# handle the case where the symbol can be found
# right at the start of the file separately
if pos == 0:
    endPos = data.find(" ")
    if endPos == -1:
        completes.append(data)
        pos = -1
    else:
        completes.append(data[:endPos])
        pos = data.find(prefix, endPos)
# pos != 0
prefix = f" {prefix}"
while pos != -1:
    pos += 1
    endPos = line.find(" ", pos)
    if endPos == -1:
        completes.append(line[pos:])
        pos = -1
    else:
        completes.append(line[pos:endPos])
        pos = line.find(prefix, endPos)
```

simonlindholm: Huh, apparently the `(?: |^)` pattern is super slow. Using ` ` instead makes it faster, but the last idea with manual `.find()` calls is faster still. Let's go with that.

Map files seem to start off as
```

Discarded input sections

```
(and also not have newlines just before symbols, whew!), so we can drop the `data.startswith(prefix)` case. It seems like symbols are newline-terminated, should we perhaps do
```py
endPos = data.find("\n", pos)
symbol = data[pos:endPos]
if " " not in symbol:
  completes.append(symbol)
```
?

Also, interestingly, my times for the non-regex methods are 10x faster than yours.

Dragorn421: <details>

<summary>symbolcompletespeedtest.py (click to expand)</summary>

```py
import time
import re

warmUpUntil = time.time() + 1
while time.time() < warmUpUntil:
    time.sleep(0.001)

n = 20
mapFile = 'z64.map'
prefix = 'EnBox'

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        for line in f:
            pos = line.find(prefix)
            # if found the prefix at start of a word
            # (start of line or if there's a space before)
            if pos >= 0 and (pos == 0 or line[pos-1] == " "):
                symbolEndPos = line.find(" ", pos)
                if symbolEndPos == -1:
                    # -1 strips the line return '\n' (assuming Unix)
                    symbol = line[pos:-1]
                else:
                    symbol = line[pos:symbolEndPos]
                completes.append(symbol)

end = time.time()

print('find/slice > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        for line in f:
            pos = line.find(prefix)
            # if found the prefix at start of a word
            # (start of line or if there's a space before)
            if pos >= 0 and (pos == 0 or line[pos-1] == " "):
                completes.append(line[pos:].split()[0] or "")

end = time.time()

print('find/split > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        for line in f:
            pass

end = time.time()

print('read file per-line > time per read:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        f.read()

end = time.time()

print('read file f.read() > time per read:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r'(?: |^)([^ ]+)(?: |\n)')
    completes = []
    with open(mapFile) as f:
        for line in f:
            symbols = prog.findall(line)
            completes.extend(symbol for symbol in symbols if symbol.startswith(prefix))

end = time.time()

print('regex findall symbols then filter > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        data = f.read()
        completes = [m.group(1) for m in re.finditer(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)", data, re.MULTILINE)]

end = time.time()

print('regex finditer prefix-in-regex > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)")
    completes = []
    with open(mapFile) as f:
        data = f.read()
        completes.extend(m.group(1) for m in prog.finditer(data, re.MULTILINE))

end = time.time()

print('regex finditer prefix-in-regex compile, extend (not much different) > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    prog = re.compile(r"(?: |^)" + re.escape(prefix) + r"([^ \n]*)(?: |$)")
    completes = []
    with open(mapFile) as f:
        for l in f:
            completes.extend(m.group(1) for m in prog.finditer(l))

end = time.time()

print('regex finditer prefix-in-regex per-line > time per search:', (end - start) / n * 1000, 'ms')

start = time.time()

for i in range(n):
    completes = []
    with open(mapFile) as f:
        data = f.read()
        if data.startswith(prefix):
            endPos = data.find(" ")
            completes.append(data if endPos == -1 else data[:endPos])
        search = f" {prefix}"
        pos = data.find(search)
        while pos != -1:
            # skip the space character in the search string
            pos += 1
            endPos = data.find(" ", pos)
            if endPos == -1:
                match = data[pos:]
                pos = -1
            else:
                match = data[pos:endPos]
                pos = data.find(search, endPos)
            completes.append(match.strip())

end = time.time()

print('find/slice f.read() > time per search:', (end - start) / n * 1000, 'ms')
```

</details>

```
find/slice > time per search: 395.8820939064026 ms
find/split > time per search: 396.27147912979126 ms
read file per-line > time per read: 348.4336853027344 ms
read file f.read() > time per read: 56.88859224319458 ms
regex findall symbols then filter > time per search: 990.7907366752625 ms
regex finditer prefix-in-regex > time per search: 487.98606395721436 ms
regex finditer prefix-in-regex compile, extend (not much different) > time per search: 536.2620830535889 ms
regex finditer prefix-in-regex per-line > time per search: 1023.4032869338989 ms
find/slice f.read() > time per search: 64.10245895385742 ms
```

I didn't think f.read() was that much faster compared to iterating over lines

`finditer` is quite slow as well but I eventually came up with this which seems to perform best (15% overhead compared to a simple f.read()):

```py
data = f.read()
if data.startswith(prefix):
    endPos = data.find(" ")
    completes.append(data if endPos == -1 else data[:endPos])
search = f" {prefix}"
pos = data.find(search)
while pos != -1:
    # skip the space character in the search string
    pos += 1
    endPos = data.find(" ", pos)
    if endPos == -1:
        match = data[pos:]
        pos = -1
    else:
        match = data[pos:endPos]
        pos = data.find(search, endPos)
    completes.append(match.strip())
```

diff.py
```diff
@@ -26,30 +26,34 @@ def fail(msg):
...
-                    if symbolEndPos == -1:
-                        # -1 strips the line return '\n' (assuming Unix)
-                        symbol = line[pos:-1]
-                    else:
-                        symbol = line[pos:symbolEndPos]
-                    completes.append(symbol)
+            data = f.read()
+            if data.startswith(prefix):
+                endPos = data.find(" ")
+                completes.append(data if endPos == -1 else data[:endPos])
```
Dragorn421, line 41: So getting rid of this would have no effect for map files?

diff.py
```diff
@@ -26,30 +26,34 @@ def fail(msg):
...
+            search = f" {prefix}"
```
Dragorn421, line 42: I didn't realize but this assumes a symbol always has a space before, can't be "\nSYMBOL"
I guess that's okay for linker maps since there are ton of spaces everywhere, such a big assumption does make the `data.startswith` case look foolish

Dragorn421 (commented):
I'm not satisfied with this symbol finding to be honest
You mentioned that we may be able to do something like searching for line returns, then the previous space " " character and that would delimit a symbol?

diff.py
simonlindholm: Yeah. Assuming a space before each symbol definitely seems fine to me. See also my comment here: https://github.com/simonlindholm/asm-differ/pull/31#discussion_r481094444

diff.py
```diff
@@ -1,47 +1,60 @@
...
+            data = f.read()
+            # assume symbols are prefixed by a space character
+            search = f" {prefix}"
+            pos = data.find(search)
+            while pos != -1:
+                # skip the space character in the search string
+                pos += 1
+                # assume symbols are suffixed by either a space
+                # character or a (unix-style) line return
+                endPos = min(data.find(" ", pos), data.find("\n", pos))
```
simonlindholm, line 47: nit: for the very last word, it will probably find only `"\n"` but no `" "`, so endPos will be -1, and `data[pos:]` will include the trailing newline

I don't care that much because the last word is a filename and not a symbol, but might be good to correct just for sanity's sake

simonlindholm (approved):

Dragorn421:
I noticed that tab-completing "" (empty string) of course finds about all the "words" in the map file (and it's very slow as well as useless), should I add a special case to skip reading the mapfile when the prefix is empty?

diff.py
Dragorn421: I don't see the issue? if there's a trailing newline then endPos will be the position of the \n ?

simonlindholm: `min(-1, newlinePos)` will be -1, not newlinePos

simonlindholm:
sure, skipping reading in that case seems sensible

diff.py
Dragorn421: Nice catch

PR #32, opened 2020-09-22
Add an option to ignore address differences

leoetlino:
This increases the signal-to-noise ratio of diffs in situations where
it is difficult to generate fully matching executables with all
functions and data being placed at the correct locations.

The new option only affects A64 diffs at the moment but extending it
to any other supported architecture shouldn't be too difficult.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -131,6 +131,12 @@ parser.add_argument(
     help="Pretend all large enough immediates are the same.",
 )
 parser.add_argument(
+    "-I",
+    "--ignore-addr-diffs",
+    action="store_true",
+    help="Ignore address differences. Currently only affects AArch64.",
+)
+parser.add_argument(
     "-B",
     "--no-show-branches",
     dest="show_branches",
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -822,12 +828,82 @@ def __hash__(self) -> int:
         return hash(self.key2)
 
 
+class DifferenceNormalizer:
+    def normalize(self, line: Line) -> str:
+        row = line.original
+        row = self._normalize_arch_specific(line, row)
+        row = maybe_normalize_large_imms(row)
```
simonlindholm, line 835: inline maybe_normalize_large_imms here and remove it, since this is the only usage

simonlindholm (approved):
Nice!

diff.py
```diff
@@ -822,12 +828,82 @@ def __hash__(self) -> int:
...
+        that are part of an ADRP + LDR/ADD pair. Registers are removed from the set as soon
+        as they are used for an LDR or ADD instruction which completes the pair.
+
+        This method is somewhat crude but should manage to detect most such pairs.
+        """
+        row_parts = row.split("\t", 1)
+        if line.mnemonic == "adrp":
+            self._adrp_pair_registers.add(row_parts[1].strip().split(",")[0])
+            row, _ = split_off_branch(row)
+            row = normalize_imms(row) + " <imm>"
```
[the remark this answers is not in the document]
leoetlino: Oh whoops... Good catch, `row` cannot have any immediates left in it after split_off_branch. I've removed this entire line as it is useless.

diff.py
```diff
@@ -865,9 +935,7 @@ def do_diff(basedump: str, mydump: str) -> List[OutputLine]:
             line_color1 = line_color2 = sym_color = Fore.RESET
             line_prefix = " "
             if line1 and line2 and line1.diff_row == line2.diff_row:
-                if maybe_normalize_large_imms(
-                    line1.original
-                ) == maybe_normalize_large_imms(line2.original):
+                if normalizer1.normalize(line1) == normalizer2.normalize(line2):
```
simonlindholm, line 938: hm, another thing. The normalizer is stateful, but that's not at all clear from this caller. I could see accidentally adding more calls to normalize(), and even now normalize() is only sometimes called for each line, while it probably should always be. Should this move into run_objdump maybe, adding another member to Line?

leoetlino: Right, that would be less error prone. I've added a "normalized_original" attribute to Line and moved the normalizer to `process`

PR #33, opened 2020-09-27
Adding a score to the diff

FryJay:


FryJay:
Doesn't work when the function is the last one in the file so I'll need to look into that.

simonlindholm:
Closing in favor of #51.

PR #39, opened 2021-01-25
Add option to use a custom make target

leoetlino:
For some build systems, the name of the target that generates the main
executable does not match the path to the ELF file, especially when
the ELF is in a separate build directory.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1533,6 +1533,10 @@ def main() -> None:
     else:
         make_target, basecmd, mycmd = dump_binary()
 
+    map_build_target_fn = getattr(diff_settings, "map_build_target", None)
+    if map_build_target_fn:
+        make_target = map_build_target_fn(make_target=make_target)
+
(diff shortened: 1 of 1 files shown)


simonlindholm:
Wouldn't it make more sense to put this in diff_settings.py? Or can it vary between invocations?

leoetlino:
If this is only used in ELF diffing mode, it would indeed make more sense to make this a setting, but this would be less flexible in general. For example, if object files are being diffed and we only want to rebuild one object file, the target name would need to vary.

simonlindholm:
Would it be possible to have diff_settings.py contain a function that maps the object file to a build target, in that case?

(I'm not opposed to taking this, per se, it just feels like it should be unnecessary)

leoetlino:
Oh yeah, that's probably a better idea :+1:

leoetlino:
Replaced it with an optional `map_build_target` function in diff_settings

PR #40, opened 2021-03-25
Clear Windows Terminal buffer before diffing

Kelebek1:
Windows Terminal doesn't clear its buffers properly and scrolling is all screwy. Diffs don't open with new buffers, and it's possible to scroll up to view previous console output. This workaround forces a clear which clears that out. If there's a better way/place to do this, let me know.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1446,6 +1446,12 @@ class Display:
             header_lines = [header] if header else []
             output = "\n".join(header_lines + diff_lines[args.skip_lines :])
 
+        # Windows Terminal does not handle buffers properly
+        # Janky hack to clear its scrollback buffer until it's fixed
+        clear_proc = subprocess.Popen(
+            ["echo", "-en", "\"\e[3J\""]
+        )
+
         # Pipe the output through 'tail' and only then to less, to ensure the
         # write call doesn't block. ('tail' has to buffer all its input before
         # it starts writing.) This also means we don't have to deal with pipe


diff.py
```diff
@@ -1446,6 +1446,12 @@ def run_less(self) -> "Tuple[subprocess.Popen[bytes], subprocess.Popen[bytes]]":
             header_lines = [header] if header else []
             output = "\n".join(header_lines + diff_lines[args.skip_lines :])
 
+        # Windows Terminal does not handle buffers properly
+        # Janky hack to clear its scrollback buffer until it's fixed
+        clear_proc = subprocess.Popen(
+            ["echo", "-en", "\"\e[3J\""]
+        )
```
simonlindholm, line 1453: Popen() starts a new process that runs asynchronously; we'd need to wait for it to finish if we do this (clear_proc.wait() or subprocess.run()/similar). Does it work if you instead do
```py
        sys.stdout.write("\x1b7[3J\"")
        sys.stdout.flush()
```
?

There is a `os.system("tput reset")` further down which is similar to this btw, but I guess it doesn't help for whatever reason (and it doesn't do anything for the very first less call).

simonlindholm (changes requested):

diff.py
Kelebek1: The stdout writes don't work unfortunately.

PR #41, opened 2021-07-13
Add support for HTML output

zbanks:
Add the `--format` argument for specifying the output style/format:
- `color` is the default option, the existing state
- `plain` also uses the terminal, but without any ANSI escape codes/styles
- `html` outputs an HTML document fragment (and disables the pager)

The stylesheet I have is pretty bad, but it should be very easy to plug in different styles (e.g. dark mode). It's mostly a 1-1 translation, but the colors should probably be tweaked for a light background.

In a future PR, it'd be nice if everywhere `COLOR_ROTATION` formatting is used, we could mark up the HTML with metadata about the symbol names. This could make it possible to add interactivity with JS?

- [x] ~Todo~: need to look closely to ensure that all untrusted strings are escaped
- [x] ~Todo~: test with `--elf`/`--source` (is there a project where these work?)

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -540,7 +540,14 @@ def search_map_file(
             return cands[0]
     elif project.map_format == "mw":
         #                                         ram   elf rom                                                       object name
-        find = re.findall(re.compile(r'  \S+ \S+ (\S+) (\S+)  . ' + fn_name + r'(?: \(entry of \.(?:init|text)\))? \t(\S+)'), contents)
+        find = re.findall(
+            re.compile(
+                r"  \S+ \S+ (\S+) (\S+)  . "
+                + fn_name
+                + r"(?: \(entry of \.(?:init|text)\))? \t(\S+)"
+            ),
+            contents,
+        )
         if len(find) > 1:
             fail(f"Found multiple occurrences of function {fn_name} in map file.")
         if len(find) == 1:
@@ -556,7 +563,9 @@ def search_map_file(
             ]
             if len(objfiles) > 1:
                 all_objects = "\n".join(objfiles)
-                fail(f"Found multiple objects of the same name {objname} in {project.mw_build_dir}, cannot determine which to diff against: \n{all_objects}")
+                fail(
+                    f"Found multiple objects of the same name {objname} in {project.mw_build_dir}, cannot determine which to diff against: \n{all_objects}"
+                )
             if len(objfiles) == 1:
                 objfile = objfiles[0]
                 # TODO Currently the ram-rom conversion only works for diffing ELF
@@ -601,7 +610,11 @@ def dump_elf(
     return (
         project.myimg,
         (objdump_flags + flags1, project.baseimg, None),
-        (objdump_flags + flags2 + maybe_get_objdump_source_flags(config), project.myimg, None),
+        (
+            objdump_flags + flags2 + maybe_get_objdump_source_flags(config),
+            project.myimg,
+            None,
+        ),
     )
 
 
@@ -1293,12 +1306,21 @@ def do_diff(basedump: str, mydump: str, config: Config) -> List[OutputLine]:
                     out2, branch2 = split_off_branch(line2.original)
                 branchless1 = out1
                 branchless2 = out2
-                out1, out2 = color_fields(arch.re_imm, out1, out2, lambda s: f"{Fore.LIGHTBLUE_EX}{s}{Style.RESET_ALL}")
+                out1, out2 = color_fields(
+                    arch.re_imm,
+                    out1,
+                    out2,
+                    lambda s: f"{Fore.LIGHTBLUE_EX}{s}{Style.RESET_ALL}",
+                )
 
                 same_relative_target = False
                 if line1.branch_target is not None and line2.branch_target is not None:
-                    relative_target1 = eval_line_num(line1.branch_target) - eval_line_num(line1.line_num)
-                    relative_target2 = eval_line_num(line2.branch_target) - eval_line_num(line2.line_num)
+                    relative_target1 = eval_line_num(
+                        line1.branch_target
+                    ) - eval_line_num(line1.line_num)
+                    relative_target2 = eval_line_num(
+                        line2.branch_target
+                    ) - eval_line_num(line2.line_num)
                     same_relative_target = relative_target1 == relative_target2
 
                 if not same_relative_target:
@@ -1306,14 +1328,20 @@ def do_diff(basedump: str, mydump: str, config: Config) -> List[OutputLine]:
 
                 out1 += branch1
                 out2 += branch2
-                if normalize_imms(branchless1, arch) == normalize_imms(branchless2, arch):
+                if normalize_imms(branchless1, arch) == normalize_imms(
+                    branchless2, arch
+                ):
                     if not same_relative_target:
                         # only imms differences
                         sym_color = Fore.LIGHTBLUE_EX
                         line_prefix = "i"
                 else:
-                    out1, out2 = color_fields(arch.re_sprel, out1, out2, sc3.color_symbol, sc4.color_symbol)
-                    if normalize_stack(branchless1, arch) == normalize_stack(branchless2, arch):
+                    out1, out2 = color_fields(
+                        arch.re_sprel, out1, out2, sc3.color_symbol, sc4.color_symbol
+                    )
+                    if normalize_stack(branchless1, arch) == normalize_stack(
+                        branchless2, arch
+                    ):
                         # only stack differences (luckily stack and imm
                         # differences can't be combined in MIPS, so we
                         # don't have to think about that case)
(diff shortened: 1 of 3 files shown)


diff.py
```diff
@@ -335,6 +351,16 @@ def create_project_settings(settings: Dict[str, Any]) -> ProjectSettings:
...
 def create_config(args: argparse.Namespace, project: ProjectSettings) -> Config:
+    formatter: Formatter
+    if args.format == "plain":
+        formatter = PlainFormatter(column_width=args.column_width)
+    elif args.format == "color":
+        formatter = AnsiFormatter(column_width=args.column_width)
+    elif args.format == "html":
+        formatter = HtmlFormatter()
+    else:
+        raise ValueError("Unsupported --format: {args.format}")
```
simonlindholm, line 362:
```suggestion
        raise ValueError(f"Unsupported --format: {args.format}")
```

diff.py
```diff
@@ -414,6 +429,251 @@ def get_arch(arch_str: str) -> "ArchSettings":
...
+            output += "    <tr>"
+            for cell in line:
+                output += f"<td>{cell}</td>"
+            output += "</tr>\n"
+        output += "</table>\n"
+        return output
+
+    @staticmethod
+    def escape(s: str) -> str:
+        # TODO: Check that all untrusted strings are passed through this
```
simonlindholm, line 673: Can we make an Html class, with "trusted input" and "text" static constructors? Or maybe it should be FormattedText, with appropriate constructors within each Formatter. (I wouldn't mind "escaping" within the AnsiFormatter by changing escape codes to unicode replacement chars, not that there's actually a case for it given it runs on trusted input.)

diff.py
```diff
@@ -414,6 +429,251 @@ def get_arch(arch_str: str) -> "ArchSettings":
...
+            return line
+        class_name = f.name.lower().replace("_", "-")
+        return f"<span class='{class_name}'>{line}</span>"
+
+    def table(
+        self, header: Optional[Tuple[str, ...]], lines: List[Tuple[str, ...]]
+    ) -> str:
+        if header:
+            lines = [header] + lines
+        # TODO: Make a prettier stylesheet (maybe include it externally)
```
simonlindholm, line 593: We shouldn't force a particular stylesheet on consumers of this, which may clash with their existing designs and be impossible to modify. An external file may be okay as long as we don't emit a `<link rel="stylesheet">` for it. But I wouldn't mind leaving this out of the repo until we make a usable html diff tool that's part of this repo, if we ever do. (I imagine such a tool would support watch mode using websockets or something)

simonlindholm (commented):
> In a future PR, it'd be nice if everywhere COLOR_ROTATION formatting is used, we could mark up the HTML with metadata about the symbol names

Agreed. It's kinda weird that COLOR_ROTATION is a thing outside AnsiFormatter. In particular it would be nice to allow the HTML formatter to use more colors; it does happen regularly that colors get reused in confusing ways, especially for branch arrows.

> Todo: test with --elf/--source (is there a project where these work?)

botw

diff.py
zbanks: I really like that pattern, but it was very difficult to incorporate into `do_diff()`. That function repeatedly re-formats the same string, which was wreaking havoc on trying to track provenance. 

After sleeping on it, I think I have two ideas for making it work -- I'll try today.

diff.py
zbanks: Yeah 100% agreed.

diff.py
zbanks: I implemented a `Text` class that tracks a list of `(plaintext, formatting)` tuples that are preserved all the way to `OutputLine`. They aren't reified into `str`s until right before they're output.

lmk what you think of the API. (Maybe there's an argument that `Formatter.table(...)` should operate on `Text`s instead of `str`s? I'm less worried about that boundary because it's only used in 1 place.)

diff.py
```diff
@@ -556,10 +572,14 @@ def table(
 class AnsiFormatter(Formatter):
     column_width: int
 
-    def apply(self, line: str, f: Format) -> str:
-        if f == Format.NONE:
-            return line
-        return f"{f.ansi_code}{line}{Style.RESET_ALL}"
+    def apply(self, text: Text) -> str:
```
simonlindholm, line 575: maybe cleaner to have this receive `chunk` and `f` and put the loop in the caller

diff.py
```diff
@@ -556,10 +572,14 @@ def table(
...
+        output = ""
+        for chunk, f in text.segments:
+            if f == Format.NONE:
+                output += chunk
+            else:
+                output += f"{f.ansi_code}{chunk}{Style.RESET_ALL}"
+        return output
```
simonlindholm, line 582: I think it's preferable to make `output` a `List[str]` and `return "".join(output)` to avoid theoretical O(n^2) performance from O(n) string concatenation... though it's not a big deal since CPython in practice optimizes `x += y` where `x` has refcount = 1.

diff.py
```diff
@@ -475,10 +476,71 @@ def __init__(self, value: int, ansi_code: str):
...
+    def formatted(line: str, f: Format) -> "Text":
+        return Text([(line, f)])
+
+    @staticmethod
+    def raw(line: str) -> "Text":
+        return Text.formatted(line, Format.NONE)
+
+    @staticmethod
+    def empty() -> "Text":
+        return Text([])
```
simonlindholm, line 493: having a single constructor `def __init__(self, line: Optional[str] = None, f: Format = Format.NONE)` might look nicer

diff.py
```diff
@@ -475,10 +476,71 @@ def __init__(self, value: int, ansi_code: str):
...
+            i = 0
+            for match in pat.finditer(chunk):
+                start, end = match.start(), match.end()
+                assert i <= start <= end <= len(chunk)
+                sub = sub_fn(match)
+                new_segments.append((chunk[i:start], f))
+                new_segments.extend(sub.segments)
+                i = end
+            new_segments.append((chunk[i:], f))
+        return Text(new_segments)
```
simonlindholm, line 537: Heh, it's funny how this illuminates how much sketchy stuff we were doing on formatted text. It would be nice to refactor the code to only do this kind of processing on `str`'s, and also get rid of a bunch of `.plain()` and `.reformat()` in the process. But I totally understand if you don't want to dig into that mess.

diff.py
simonlindholm: That sounds like a nice API to me!

zbanks: > That sounds like a nice API to me!

Just checking - do you mean the `Text` API as-implemented, or are you saying I *should* also change `Formatter.table(...)` to work with `Text`s in this PR?

simonlindholm: the `Text` API as implemented. I wouldn't mind changing Formatter.table too (would get rid of the need for ansi_ljust which is nice) but as you say it's not particularly worrying

diff.py
zbanks: Yeah, I think a nicer approach would involve tokenizing the lines first? And then using the regexes to match tokens, rather than using `re.sub(...)`-style operations. 
Tokenizing might be easy -- it be a simple as splitting on `\t` and `,`? But overall that seemed like a bigger change and I was nervous about getting it right for all 3 supported arches.

diff.py
zbanks: Yeah I think this makes it *a lot* nicer! Good call!

diff.py
simonlindholm: Tokenizing would be nice for sure, but yeah, it's a larger change and it may be a good idea to build up a small test suite first... What I was hoping for was more of a smaller change that keeps `out1`, `out2` as `str` for long enough that regexes-on-Text wouldn't be necessary

diff.py
```diff
@@ -1540,8 +1536,8 @@ def do_diff(basedump: str, mydump: str, config: Config) -> List[OutputLine]:
     for (line1, line2) in diff_lines(lines1, lines2, config.algorithm):
         line_color1 = line_color2 = sym_color = Format.NONE
         line_prefix = " "
-        out1 = Text.empty() if not line1 else Text.raw(pad_mnemonic(line1.original))
-        out2 = Text.empty() if not line2 else Text.raw(pad_mnemonic(line2.original))
+        out1 = Text() if not line1 else Text(pad_mnemonic(line1.original))
+        out2 = Text() if not line2 else Text(pad_mnemonic(line2.original))
```
simonlindholm, line 1540: beware 8ffce9b63939a7329db3670c768a46360ea32b72 if you're moving `pad_mnemonic` early. I don't remember the details of that unfortunately so I can't say if it applies

zbanks: Seems like should be fine now, because it only adds whitespace to the mnemonic term (before it also padded the immediate)?

diff.py
```diff
@@ -414,6 +430,240 @@ def get_arch(arch_str: str) -> "ArchSettings":
...
+
+    def __eq__(self, other: object) -> bool:
+        return NotImplemented
+
+    def __add__(self, other: Union["Text", str]) -> "Text":
+        if isinstance(other, str):
+            return self + Text(other)
+        result = Text()
+        result.segments.extend(self.segments)
+        result.segments.extend(other.segments)
```
simonlindholm, line 510:
```suggestion
        result.segments = self.segments + other.segments
```

simonlindholm (approved):

diff.py
```diff
@@ -414,6 +430,240 @@ def get_arch(arch_str: str) -> "ArchSettings":
...
+        if f == Format.NONE:
+            return chunk
+        class_name = f.name.lower().replace("_", "-")
+        return f"<span class='{class_name}'>{chunk}</span>"
+
+    def table(
+        self, header: Optional[Tuple[str, ...]], lines: List[Tuple[str, ...]]
+    ) -> str:
+        if header:
+            lines = [header] + lines
```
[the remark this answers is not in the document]
zbanks: Oh yeah - I was intending to do that, I'll fix that now.

PR #42, opened 2021-07-15
Expose highlighted branch/reg details in an HTML attribute

zbanks:
This PR doesn't change the visible output, but should make it straightforward to improve the HTML diff. Should be easier to add new colors, incorporate `<a>`s, or write JS to make it more interactive.

I don't love having separate `BasicFormat`/`RotationFormat` types, but I couldn't find a better way to emulate a sum type with modern Python. (Happy to hear a better suggestion!)

I also think the naming/format of the `data-rotation` is a bit arbitrary, I mostly wanted to make sure that the information was plumbed through without being excessively verbose.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -24,6 +24,10 @@ def fail(msg: str) -> NoReturn:
     sys.exit(1)
 
 
+def static_assert_unreachable(x: NoReturn) -> NoReturn:
+    raise Exception("Unreachable! " + repr(x))
+
+
 # Prefer to use diff_settings.py from the current working directory
 sys.path.insert(0, ".")
 try:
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -599,12 +623,22 @@ def ansi_ljust(self, s: str) -> str:
 
 @dataclass
 class HtmlFormatter(Formatter):
+    rotation_formats: int = 9
```
simonlindholm, line 626: I guess it's fine for an embedder for set `HtmlFormatter.rotation_formats = something else`

simonlindholm (approved):

PR #43, opened 2021-07-16
Fix --watch not working with files in top directory

Dragorn421:
I'm tinkering with diff.py and for testing I'm using target.bin and source.bin files in the same directory as diff.py (running with `./diff.py --watch 0`)
--watch doesn't see changes made to source.bin because `WatchEventHandler#changed` receives "./source.bin" which isn't in the watched files list `file_targets == ['source.bin']`
I fixed it by changing `if path == target:` to `if os.path.samefile(path, target):` in `WatchEventHandler#should_notify`, I thought that was the cleanest

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1737,7 +1737,7 @@ def debounced_fs_watch(
 
         def should_notify(self, path: str) -> bool:
             for target in self.file_targets:
-                if path == target:
+                if os.path.samefile(path, target):
                     return True
             if config.make and any(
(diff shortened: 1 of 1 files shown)


simonlindholm:
os.path.samefile involves a syscall, which could be a bit slow. How about calling `os.path.normpath` on all path names instead?

Dragorn421:
I don't think speed is an issue here, but sure!

PR #44, opened 2021-07-17
Add --source-old-binutils, like --source but works with binutils < 2.33

Dragorn421:
Using `--source` with diff.py in main relies on the objdump `--source-comment` option which apparently was only added in binutils 2.33+

This adds `--source-old-binutils` which makes do without `--source-comment` at the cost of using a regex to match source lines (hopefully it won't do (many) mistakes) to achieve what `--source` does with binutils 2.33+

I also added a rough check for objdump failing because of `--source-comment` to suggest using `--source-old-binutils`

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -793,7 +793,7 @@ def run_objdump(cmd: ObjdumpCommand, config: Config, project: ProjectSettings) -
         print(e.stdout)
         print(e.stderr)
         if "unrecognized option '--source-comment" in e.stderr:
-            print("** Try using --source-old-binutils instead of --source **")
+            fail("** Try using --source-old-binutils instead of --source **")
         raise e
 
     if restrict is not None:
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -774,10 +783,19 @@ def maybe_get_objdump_source_flags(config: Config) -> List[str]:
...
+            [project.objdump_executable] + config.arch.arch_flags + flags + [target],
+            check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+            universal_newlines=True,
+        ).stdout
+    except subprocess.CalledProcessError as e:
+        print(e.stdout)
+        print(e.stderr)
+        if "unrecognized option '--source-comment" in e.stderr:
+            print("** Try using --source-old-binutils instead of --source **")
+        raise e
```
simonlindholm, line 797: maybe sys.exit(1) to avoid the suggestion getting lost in error spew noise?

simonlindholm (approved):
Nice, I like this approach

PR #45, opened 2021-07-20
View diff in browser

Dragorn421:
### Command line arguments

Added `--no-make` to bypass the warning when using `--watch` without `--make`

Added `--web` to make diff.py serve diffs in html format over http

Added `--web-port` for setting the http server port (default 8000)

Added `--browse` to run the browser automatically (when using `--web`) (uses python's `webbrowser.open` or, if set, the `ASMDW_BROWSER_CMD` environment variable (eg `export ASMDW_BROWSER_CMD='firefox.exe http://localhost:{port}?{query}'`))

### css/html

Changed colors in `diff-stylesheet.css` to match a "dark blue theme" and fix contrast issues

Added client.html as the client (diff.py being the server), and a script.js. Renamed `diff-stylesheet.css` to style.css and move all of that to a new web/ folder

### Display

Finally, I refactored the Display class by moving most of it to PagerDisplay and making Display an abstract class that is now also implemented by ConsoleDisplay (`--no-pager` without `--web`) and WebDisplay (`--web`).

Example with `./diff.py -mwo3 --web --browse func_80B42F74` (OoT decomp):

<details>

</details>

### Plans

Get your opinion on the state of this, what's left to clean up

I think I want to add UI controls in-browser like a "find first diff" button and other stuff

I don't think it is worth implementing a UI for giving directives to diff.py like changing the diffed function since it's easy to do that from CLI, opinion?


simonlindholm:
This is super cool, will fix some long-standing issues like flickerless updates and maintaining scroll position if it becomes the main workflow, and I'm hyped for the interactivity features it allows (easy screenshotting, highlighting registers and branches, extensibility with project-specific tools, ...).

> I think I want to add UI controls in-browser like a "find first diff" button and other stuff

That sounds really neat. I guess it should be part of diff_settings.py, so that projects can add their own custom tools?

> I don't think it is worth implementing a UI for giving directives to diff.py like changing the diffed function since it's easy to do that from CLI, opinion?

It could be nice to allow a non-cmdline workflow for decomp. Changing the threeway diff base is another useful action. I think we'll see how it feels after we get more experience with using this, hard to say right now whether it's worth the effort.

diff-stylesheet.css
```diff
@@ -1,64 +1,62 @@
 table.diff {
-    border: none;
-    font-family: Monospace;
-    white-space: pre;
+	border: none;
+	font-family: Monospace;
+	white-space: pre;
+	background-color: #243341;
+	color: #ecf0f1;
```
simonlindholm, line 6: I think we should use spaces (also for the HTML)

diff.py
```diff
@@ -200,6 +200,11 @@ def complete_symbol(
         help="Automatically update when source/object files change. "
         "Recommended in combination with -m.",
     )
+    parser.add_argument(
+        "--no-make",
```
simonlindholm, line 204: `dest="no_make"` to make this greppable. Alternatively use `dest="make" action="store_const" const=False` (and const=True for --make) and check for whether make is True, False or None

diff.py
```diff
@@ -352,11 +375,14 @@ def create_project_settings(settings: Dict[str, Any]) -> ProjectSettings:
         objdump_executable=get_objdump_executable(settings.get("objdump_executable")),
         map_format=settings.get("map_format", "gnu"),
         mw_build_dir=settings.get("mw_build_dir", "build/"),
+        browser=settings.get("browser"),
```
simonlindholm, line 378: I don't think it's right to have this in project settings, it's a personal preference. Use an env var instead. There's also `webbrowser.open` in Python which could serve as a default, though I'm not sure how well it works with WSL or if it's worth the complexity to support (also makes it impossible to give an error message about what kind of env var to create)

diff.py
```diff
@@ -1880,21 +1950,136 @@ def progress(self, msg: str) -> None:
...
+
+    def run_sync(self) -> None:
+        web_display = self
+
+        class MyReqHandler(http.server.BaseHTTPRequestHandler):
+            def do_GET(self):
+                if not web_display.running:
+                    return
+                self.send_response(http.HTTPStatus.OK)
+                self.send_header("Access-Control-Allow-Origin", "*")
```
simonlindholm, line 1983: (seems sketchy)

diff.py
```diff
@@ -1880,21 +1950,136 @@ def progress(self, msg: str) -> None:
...
+                # fixme
+                threading.Thread(target=web_display.http_server.shutdown).start()
+
+        self.http_server = http.server.HTTPServer(
+            ("localhost", self.config.http_server_port), MyReqHandler
+        )
+        self.running = True
+        # fixme
+        print(f"Http server is running on port {self.config.http_server_port}")
+        print("Open with ?once (eg file:///.../client.html?once)")
```
simonlindholm, line 1996: file:/// isn't great; it's generally low-privileged and behaves differently between browsers wrt what files it can access. would be better if we could serve it from the same port

diff.py
```diff
@@ -1880,21 +1950,136 @@ def progress(self, msg: str) -> None:
...
+    def run_sync(self) -> None:
+        self.print_diff()
+
+    def run_async(self, watch_queue: "queue.Queue[Optional[float]]") -> None:
+        self.print_diff()
+
+    def progress(self, msg: str) -> None:
+        print(msg)
+
+    def update(self, mydump: str) -> None:
```
simonlindholm, line 2070: is this ever run? seems weird to have watch but no pager

client.html
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		function onBodyLoaded() {
+			requestContent();
+		}
+
+		// asm-differ web
+		window.ASMDW = {};
+		var url = new URL(window.location);
+		ASMDW.requestForever = url.searchParams.get('once') === null;
```
simonlindholm, line 42: not a fan of needing the `once` param, should ideally be part of the response instead. Potentially as a header if it's hard to fit in the body. Alternatively we can let the JS long-poll and just never get an update, which seems harmless and may make the code simpler, and would work better with the "red status marker" idea mentioned above

client.html
```diff
@@ -0,0 +1,55 @@
...
+			var httpRequest = ASMDW.httpRequest;
+			if (httpRequest.readyState === XMLHttpRequest.DONE) {
+				if (httpRequest.status === 200) {
+					document.getElementById('content').innerHTML = httpRequest.responseText;
+					if (ASMDW.requestForever) {
+						setTimeout(requestContent, 100);
+					}
+				}
+				else {
+					if (window.confirm('Failed to communicate with localhost:8000, close tab?')) {
```
simonlindholm, line 20: it may be nice to show a red status marker instead, and keep polling (with exponential backoff?). Then you won't get a prompt every time you diff.py a new function, and can instead continue using the same tab. (But this is fine for now ofc)

diff.py
Dragorn421: I'll go with `dest="no_make"`, sounds less spaghetti :p

diff.py
Dragorn421: `Access-Control-Allow-Origin: *` ? The header is needed for the browser to allow javascript to make the request

I could use `null` instead of `*`, at least on firefox, not sure if it's standard for file:// origins to have `Origin: null`

I could also use JSONP (the web technology has incredibly bad names) to not deal with CORS but that's way more sketchy imo

diff.py
Dragorn421: You mean serve client.html through diff.py ? good idea

diff.py
Dragorn421: it is and it works, if you run --watch --no-pager you get the diff spit out in stdout. Not usable by any mean but if you want it it's there
the main code (that does the refreshing if --watch) is Display-subclass-independent, if that makes sense

client.html
Dragorn421: Yeah about this and the last comment (about red status marker) I just need to improve on how client-server transactions are made, it's currently not displaying status such as "Building" or "Unchanged" in-browser either

diff.py
Dragorn421: For some reason `webbrowser.open` run from the python3 prompt in WSL does open me a firefox tab, neat. Even more impressive (imo), `webbrowser.open('client.html')` can open client.html from the wsl filesystem (`file://///wsl$/Ubuntu-18.04/home/dragorn421/asm-differ/client.html`)!
Though as discussed, the open url will eventually be "http://localhost:8000" or smth like that

So I used both an env var and webbrowser.open (env var preferred)

client.html
```diff
@@ -53,6 +78,221 @@
...
+                            // branch origins have data-branch-target
+                            if ('branchTarget' in spanChild.dataset) {
+                                spanChild.addEventListener('click', onClickBranchOrigin);
+                            } else {
+                                spanChild.addEventListener('click', onClickBranchTarget);
+                            }
+                            spanChild.addEventListener('mouseenter', onMouseEnterBranchIndicator);
+                            spanChild.addEventListener('mouseleave', onMouseLeaveBranchIndicator);
+                            spanChild.addEventListener('animationend',
+                                ev => ev.target.classList.remove('highlight-branch'));
```
simonlindholm, line 263: instead of setting event listeners on individual elements like this, I'd suggest setting a single one on document and having it check what to do depending on event.target

client.html
```diff
@@ -53,6 +78,221 @@
...
+        }
+
+        function setDiffHtml(diffHtml) {
+            ASMDW.dom.diffContainer.innerHTML = diffHtml;
+            var tbody = ASMDW.dom.diffContainer.getElementsByTagName('tbody')[0];
+            for (var i = 0; i < tbody.children.length; i++) {
+                var tr = tbody.children[i];
+                for (var j = 0; j < 2 && j < tr.children.length; j++) {
+                    var td = tr.children[j];
+                    var spanChildren = td.getElementsByTagName('span');
```
simonlindholm, line 249: `diffContainer.querySelectorAll("tbody > tr > td > span")`

client.html
```diff
@@ -53,6 +78,221 @@
...
+                            wrapBranchIndicatorsElems.push(spanChild);
+                        }
+                    }
+                    for (var k = 0; k < wrapBranchIndicatorsElems.length; k++) {
+                        var wrapBranchIndicatorElem = wrapBranchIndicatorsElems[k];
+                        var wrapperElem = document.createElement('span');
+                        wrapperElem.classList.add('branch-indicator-wrapper');
+                        td.insertBefore(wrapperElem, wrapBranchIndicatorElem);
+                        wrapBranchIndicatorElem.remove();
+                        wrapperElem.appendChild(wrapBranchIndicatorElem);
```
simonlindholm, line 273: it would also be fine to modify the HTML emitted by diff.py I think

client.html
```diff
@@ -53,6 +78,221 @@
...
+            // get all branch origins that branch to this branch target (ev.target)
+            var branchIndicatorsElems = document.getElementsByClassName(ev.target.dataset.branchesClass);
+            var branchOriginsElems = [];
+            for (var i = 0; i < branchIndicatorsElems.length; i++) {
+                if (branchIndicatorsElems[i] != ev.target) {
+                    branchOriginsElems.push(branchIndicatorsElems[i]);
+                }
+            }
+
+            // use a different origin as main scroll-to element each click
```
simonlindholm, line 211: hm, I think this behavior might be undiscoverable and confusing when you trigger it by accident. may also not be necessary with persistent highlighting?

client.html
```diff
@@ -34,6 +34,28 @@
                     statusElemInfo.element.style = 'margin-top: ' + deltaY.toString() + 'px;';
                 }
             }
+            // scrolling
+            if (ASMDW.scrollYtarget !== null) {
+                var scrollYdiff = (time - ASMDW.scrollYlastUpdate) / 1000 * 1500; // 2000 px/s
```
simonlindholm: make this a constant instead of a stale comment

simonlindholm: const/let instead of var

client.html
```diff
@@ -53,6 +78,221 @@
...
+            }
+
+            ASMDW.scrollYtarget = newViewportY;
+            ASMDW.scrollYlastUpdate = performance.now();
+        }
+
+        function highlightBranchClearAll() {
+            var elems = document.getElementsByClassName('branch-indicator-wrapper');
+            for (var i = 0; i < elems.length; i++) {
+                var elem = elems[i];
```
simonlindholm, line 162: for (let elem of document.getElementsByClassName(...))

client.html
Dragorn421: how would it handle not going into the --threeway column? (in the loop I have `j<2`)
I didn't test it but I think we don't want to do anything about that column?

simonlindholm: Ah, that's true. Can add a `:not(:nth-child(3))` part to the selector

client.html
Dragorn421: done for `click`, but `mouseenter`/`mouseleave` need to be on individuals elements from what I understood

simonlindholm: It works if you use a capturing event listener (addEventListener(..., true)). May be that performance suffers

web/script.js
```diff
@@ -0,0 +1,355 @@
...
+        scrollIntoView(branchTargetElem, [elem]);
+    }
+    highlightBranchClearAll();
+    highlightBranchStay(elem.dataset.branchesClass);
+}
+
+function onClickBranchTarget(elem) {
+    // get all branch origins that branch to this branch target
+    let branchOriginsElems = [];
+    for (let branchIndicatorElem of document.getElementsByClassName(elem.dataset.branchesClass)) {
```
zbanks, line 205: I think you can simplify the HTML a bit by using `document.querySelectorAll('data-rotation="' + elem.dataset.rotation + '"')` ? (And similar, to avoid having to add many new data attributes/classes?)

Dragorn421: I'm avoiding relying on data-rotation since this "rotation" stuff is purely intended for colors rather than describing data. What I mean is I prefer verbosity over over-engineered reused data that produces hard-to-fix bugs

Also it's not like we are restrained on how much data we can send through localhost (a html diff with default max func size is ~200ko, even for decomp.me it would be manageable)

client.html
Dragorn421: Agreed, see c93db8875e27e635af66af97fce11657db652666 : I replaced this by asking the user through a popup div if the branch spots end up being too far to show all of them at once

screenshots:
<details>

How it looks with enough space:

With not enough space, clicking on the target shows:


</details>

Dragorn421:
@simonlindholm could you take a look at bc41898 

IMO it's awful but I don't really know how to greatly improve it. Specifically atm I don't see how to improve on:
1) communicating from Display to Main through watch_queue and constants looks sketchy
2) the Main class looks bad imo but I do need some kind of context to store stuff in to avoid using global variables
3) stopping the filesystem watch with a callback that's returned, and it relies on editing a list as a "local global" variable

but it works so there's that

Dragorn421:
I'm not interested in maintaining this anymore (obviously) and I guess decomp.me is enough (?), though I still think this is as-is friendlier to use than the terminal and would be a valuable addition if it fitted the repo.
But if I ever get back to this I would turn this into some kind of plug-in instead of being built-in.

Thanks for your time and thanks for decomp.me :)

PR #47, opened 2021-08-10
Add a short alias for --source (-c)

leoetlino:
After the addition of --source-old-binutils, passing --source now
requires writing the option in full. --so and other shorthands
do not work anymore because they are now ambiguous.

For the sake of convenience, this adds -c as an alias for --source
(c for sourCe or C/C++, I guess...)

~~We may want to add an alias for --source-old-binutils as well but
I don't know what alias would be best. -C? (Maybe it's best to
let people who actually use that option decide.)~~

Similarly, added -C as a alias for --source-old-binutils.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -117,11 +117,13 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--source",
+        "-c",
         action="store_true",
         help="Show source code (if possible). Only works with -o and -e.",
     )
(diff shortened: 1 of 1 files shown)


simonlindholm:
-C makes sense to me

leoetlino:
alright, I've added -C as an alias for --source-old-binutils

PR #51, opened 2021-08-30
Calculate diff score

### LINKED ISSUE #48: Expose a diff score or information such that a score could be calculated
Some options such as stack difference sensitivity would also be useful here

zbanks:
Closes #48 

- Based on `scorer.py` from decomp-permuter, but refactored to take advantage of the existing computation & format.
- Add score to header line in parentheses
    - Changed output to display header line even on two-way diffs
- Added `Diff` dataclass to bundle diff with score.
- Added configuration options to `Config`, but did not expose them.

I only tested this with MM, I don't have any of the non-MIPS projects set up locally.
The changes in this PR should be relatively easy to incorporate into #45.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -301,6 +301,7 @@ if __name__ == "__main__":
 
 import abc
 import ast
+from collections import Counter
 from dataclasses import dataclass, field, replace
 import difflib
 import enum
@@ -373,6 +374,14 @@ class Config:
     ignore_large_imms: bool
     ignore_addr_diffs: bool
     algorithm: str
+
+    # Score options
+    score_stack_differences = True
+    penalty_stackdiff = 1
+    penalty_regalloc = 5
+    penalty_reordering = 60
+    penalty_insertion = 100
+    penalty_deletion = 100
 
 
 def create_project_settings(settings: Dict[str, Any]) -> ProjectSettings:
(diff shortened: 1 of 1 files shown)


nanaian:
> Based on scorer.py from decomp-permuter, but refactored to take advantage of the existing computation & format.
Are scores consistent between decomp-permuter and this PR?

zbanks:
> > Based on scorer.py from decomp-permuter, but refactored to take advantage of the existing computation & format.
> 
> Are scores consistent between decomp-permuter and this PR?

They're not 100% one-to-one -- but they should be close. The last two commits add a heuristic to handle truncated output: the permuter has an easier job because it deals with object files containing a single function.

diff.py
```diff
@@ -1548,6 +1573,125 @@ def diff_lines(
...
+) -> int:
+    score = 0
+    deletions = []
+    insertions = []
+
+    def lo_hi_match(old: str, new: str) -> bool:
+        old_lo = old.find("%lo")
+        old_hi = old.find("%hi")
+        new_lo = new.find("%lo")
+        new_hi = new.find("%hi")
```
simonlindholm, line 1587: add a TODO comment about making this arch-independent, similar to `imm_matches_everything`

diff.py
```diff
@@ -1548,6 +1573,125 @@ def diff_lines(
     return ret
 
 
+def score_diff_lines(
+    lines: List[Tuple[Optional[Line], Optional[Line]]], config: Config
+) -> int:
```
simonlindholm, line 1578: add a comment saying this logic is copied from the permuter

diff.py
```diff
@@ -1548,6 +1573,125 @@ def diff_lines(
...
+    for index, (line1, line2) in enumerate(lines):
+        if (line1 is not None and line1.original == "...") or (
+            line2 is not None and line2.original == "..."
+        ):
+            lines_were_truncated = True
+        if line1 is not None and line2 is not None and line1.mnemonic == line2.mnemonic:
+            if start_index is None:
+                start_index = index
+            streak_len = (index - start_index) + 1
+            if streak_len >= max_len:
```
simonlindholm, line 1660: maybe we should do `streak_len >= 50` instead, to avoid a long matching streak in the middle of a large function causing the end to be cut off

diff.py
```diff
@@ -1548,6 +1573,125 @@ def diff_lines(
...
+                max_index = index
+                max_len = streak_len
+        else:
+            start_index = None
```
simonlindholm, line 1664: (`start_index = index + 1` also works and makes it possible to remove the `start_index is None` check, not sure if clearer)

diff.py
```diff
@@ -1548,6 +1573,125 @@ def diff_lines(
...
+    max_index = None
+    max_len = 0
+    start_index = None
+    lines_were_truncated = False
+    for index, (line1, line2) in enumerate(lines):
+        if (line1 is not None and line1.original == "...") or (
+            line2 is not None and line2.original == "..."
+        ):
+            lines_were_truncated = True
+        if line1 is not None and line2 is not None and line1.mnemonic == line2.mnemonic:
```
simonlindholm, line 1656: (nit: `if line1 and line2 and ...:` may be easier to read? the rest of the code is a bit inconsistent about this, I know...)

simonlindholm (approved):
I feel a bit terrified of the idea of adding `...` to the objdump and just pretending it's an instruction (it might even show up as a `|` instruction diff if one of the sides gets cut off but not the other!) ... but I think it will probably be fine?

diff.py
zbanks: I think this is what you meant?

zbanks:
> I feel a bit terrified of the idea of adding `...` to the objdump and just pretending it's an instruction (it might even show up as a `|` instruction diff if one of the sides gets cut off but not the other!) ... but I think it will probably be fine?

Heh - I think it might actually be a *good* thing if it shows up on one side but not the other? It would signal to the user that only one side was truncated -- although I agree it makes the diff a bit odd :smile:

PR #52, opened 2021-08-31
JSON output

zbanks:
- Enabled with `--format=json`, implies `--no-pager`
- JSON format should be suitable to generate HTML output
   - Output is kind of large, but gzips well (to about ~10% of original size)
- Includes score + arch as metadata
- Somewhat large refactor of `format_diff()` (now `align_diffs()`)
- [Example output](https://gist.github.com/zbanks/5c0cc3562db51cd3e965961fed575d42) from MM, indented

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -264,9 +264,9 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--format",
-        choices=("color", "plain", "html"),
+        choices=("color", "plain", "html", "json"),
         default="color",
-        help="Output format, default is color. --format=html implies --no-pager.",
+        help="Output format, default is color. --format=html or json implies --no-pager.",
     )
     parser.add_argument(
         "-U",
@@ -302,11 +302,12 @@ if __name__ == "__main__":
 import abc
 import ast
 from collections import Counter
-from dataclasses import dataclass, field, replace
+from dataclasses import asdict, dataclass, field, replace
 import difflib
 import enum
 import html
 import itertools
+import json
 import os
 import queue
 import re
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1980,73 +2079,70 @@ def flush_matching() -> None:
...
+            previous_score=old_diff.score,
+        )
+        diff_lines = [(line, line) for line in new_diff.lines]
     if config.compress:
-        ret_lines = compress_matching(diff_lines, config.compress.context)
-    else:
-        ret_lines = [line for line, _ in diff_lines]
-    return header_line, ret_lines
+        diff_lines = compress_matching(diff_lines, config.compress.context)
+    # TODO: status line, with e.g. approximate permuter score?
```
simonlindholm, line 2144:
```suggestion
```

diff.py
```diff
@@ -1980,73 +2079,70 @@ def flush_matching() -> None:
...
-            for (base, old, new) in output
-            if base or new.key2 is not None
-        ]
+        meta = TableMetadata(
+            headers=(
+                Text("TARGET"),
+                Text(f"  CURRENT ({new_diff.score})"),
+            ),
+            current_score=new_diff.score,
+            previous_score=old_diff.score,
```
simonlindholm, line 2139:
```suggestion
            previous_score=None,
```
and make previous_score nullable?

simonlindholm (approved):

diff.py
```diff
@@ -586,6 +589,24 @@ def ljust(self, column_width: int) -> "Text":
         length = sum(len(x) for x, _ in self.segments)
         return self + " " * max(column_width - length, 0)
 
+    def condense(self) -> None:
```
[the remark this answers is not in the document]
zbanks: Done. Also added some checks to `sub` to remove empty segments.

zbanks:
Not sure if you want to take another look after my `Text` refactor? Otherwise I'm ready to merge it.

simonlindholm:
Looks good to me! `sub` could create consecutive segments with the same format, but in practice I don't think we care.

PR #55, opened 2021-09-04
Show references to .text from other sections

simonlindholm:


diff.py
```diff
@@ -697,12 +700,19 @@ def apply_format(self, chunk: str, f: Format) -> str:
             ]
         else:
             static_assert_unreachable(f)
-        return f"{ansi_code}{chunk}{Style.RESET_ALL}"
+        return f"{ansi_code}{chunk}{Fore.RESET}"
```
zbanks, line 703: I think the `Style.BRIGHT` & `STYLE_UNDERLINE` also need to be disabled here? Maybe add `Style.NORMAL` and `STYLE_NO_UNDERLINE = "\u001b[24m"`?

diff.py
```diff
@@ -697,12 +700,19 @@ def apply_format(self, chunk: str, f: Format) -> str:
...
 
     def table(self, meta: TableMetadata, lines: List[Tuple["OutputLine", ...]]) -> str:
-        rows = [meta.headers] + [self.outputline_texts(ls) for ls in lines]
+        rows = [(meta.headers, False)] + [
+            (self.outputline_texts(line), line[1].is_data_ref) for line in lines
+        ]
         return "\n".join(
-            "".join(self.apply(x.ljust(self.column_width)) for x in row) for row in rows
+            "".join(
+                (Back.BLACK if is_data_ref else "")
```
zbanks, line 711: Could do invert instead of black background, so it's more likely to be readable? (`\u001b[7m` to enable; `\u001b[27m` to disable)

diff.py
```diff
@@ -1046,6 +1071,113 @@ def search_map_file(
...
+
+    is_32bit = e_ident[4] == 1
+    is_little_endian = e_ident[5] == 1
+    str_end = "<" if is_little_endian else ">"
+    str_off = "I" if is_32bit else "Q"
+    sym_size = {"B": 1, "H": 2, "I": 4, "Q": 8}
+
+    def read(spec: str, offset: int) -> Tuple[int, ...]:
+        spec = spec.replace("P", str_off)
+        size = sum(sym_size[c] for c in spec)
```
zbanks, line 1091: You can use `struct.calcsize(...)` instead. (This would also let you use things like "5H" if you wanted)

zbanks (approved):
Looks good, cool feature

I didn't examine the ELF-parsing logic in `parse_elf_data_references()` toooo closely, but it seems correct.

diff.py
simonlindholm: It looks a bit more jarring than the subtle black background:
image

but it may be fine. Let's try it, I agree it should be more portable.

PR #57, opened 2021-09-18
Add ARM32 support

abaresk:
See example [asm diff](https://gist.github.com/abaresk/bdffeae16d5764e534a81c8b88a84b30) (includes screenshot).

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -44,7 +44,7 @@ if __name__ == "__main__":
     except ModuleNotFoundError:
         argcomplete = None
 
-    parser = argparse.ArgumentParser(description="Diff MIPS, PPC or AArch64 assembly.")
+    parser = argparse.ArgumentParser(description="Diff MIPS, PPC, AArch64, or ARM32 assembly.")
 
     start_argument = parser.add_argument(
         "start",
@@ -200,7 +200,7 @@ if __name__ == "__main__":
         "--ignore-addr-diffs",
         dest="ignore_addr_diffs",
         action="store_true",
-        help="Ignore address differences. Currently only affects AArch64.",
+        help="Ignore address differences. Currently only affects AArch64 and ARM32.",
     )
     parser.add_argument(
         "-B",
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -1367,6 +1369,22 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
 
         return row
 
+class DifferenceNormalizerARM32(DifferenceNormalizer):
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
+
+    def _normalize_arch_specific(self, mnemonic: str, row: str) -> str:
+        if self.config.ignore_addr_diffs:
+            row = self._normalize_bl(mnemonic, row)
```
simonlindholm, line 1378: (would be nice to generalize this in the future to cover all archs, the logic should be the same for `jal` on mips for instance. just hasn't been a priority because for mips we almost solely use -o for which --ignore-addr-diffs isn't needed)

simonlindholm (approved):
Looks good, can you update the readme/--help/-I help texts to mention ARM32 as well?

diff.py
```diff
@@ -1477,6 +1520,20 @@ class ArchSettings:
     instructions_with_address_immediates=MIPS_BRANCH_INSTRUCTIONS.union({"jal", "j"}),
 )
 
+ARM32_SETTINGS = ArchSettings(
+    re_int=re.compile(r"[0-9]+"),
+    re_comment=re.compile(r"(<.*?>|//.*$)"),
+    # GPRs and FP registers: r0..15, pc, lr
+    # The zero registers and SP should not be in this list.
```
leoetlino, line 1527: Copy paste error? 32-bit ARM has no zero register.

diff.py
```diff
@@ -1477,6 +1520,20 @@ class ArchSettings:
...
+    re_reg=re.compile(r"\$?\b(r[0-9]|r1[0-5]|pc|lr)\b"),
+    re_sprel=re.compile(r"sp, #-?(0x[0-9a-fA-F]+|[0-9]+)\b"),
+    re_large_imm=re.compile(r"-?[1-9][0-9]{2,}|-?0x[0-9a-f]{3,}"),
+    re_imm=re.compile(r"(?<!sp, )#-?(0x[0-9a-fA-F]+|[0-9]+)\b"),
+    branch_instructions=ARM32_BRANCH_INSTRUCTIONS,
+    instructions_with_address_immediates=AARCH64_BRANCH_INSTRUCTIONS.union({"adr"}),
```
leoetlino, line 1533: ARM32_BRANCH_INSTRUCTIONS?

diff.py
```diff
@@ -1477,6 +1520,20 @@ class ArchSettings:
     instructions_with_address_immediates=MIPS_BRANCH_INSTRUCTIONS.union({"jal", "j"}),
 )
 
+ARM32_SETTINGS = ArchSettings(
+    re_int=re.compile(r"[0-9]+"),
+    re_comment=re.compile(r"(<.*?>|//.*$)"),
+    # GPRs and FP registers: r0..15, pc, lr
```
leoetlino, line 1526: There are also no VFP/NEON registers in the list at the moment.

diff.py
abaresk: Fixed.

diff.py
abaresk: Fixed.

diff.py
abaresk: Not very familiar with these. Please let me know if I specified anything incorrectly.

leoetlino: The list is not incorrect per se – it's just missing all the floating point registers. However, the comment is misleading because:

* r0..r15 includes r14, which is already mentioned separately as the link register (LR)
* r15 also appears twice (PC), and PC isn't considered a general-purpose register.
* There are no floating-point registers in the description or in the code.

abaresk: Okay, I added in the single-precision floating-point registers and tried to clarify the comment.

diff.py
abaresk: Gotcha, makes sense.

simonlindholm:
@leoetlino does this look good to you? You can merge if so

leoetlino (approved):

PR #59, opened 2021-09-19
Support mipsel for objdump

mkst:


diff.py
```diff
@@ -1544,6 +1550,10 @@ class ArchSettings:
     instructions_with_address_immediates=MIPS_BRANCH_INSTRUCTIONS.union({"jal", "j"}),
 )
 
+MIPSEL_SETTINGS = copy.deepcopy(MIPS_SETTINGS)
+MIPSEL_SETTINGS.name="mipsel"
+MIPSEL_SETTINGS.big_endian = False
```
simonlindholm, line 1555: since ArchSettings uses dataclasses you can do `MIPSEL_SETTINGS = replace(MIPS_SETTINGS, name="mipsel", big_endian=False)`

diff.py
```diff
@@ -1310,7 +1311,11 @@ def dump_binary(
         end_addr = eval_int(end, "End address must be an integer expression.")
     else:
         end_addr = start_addr + config.max_function_size_bytes
-    objdump_flags = ["-Dz", "-bbinary", "-EB"]
+    objdump_flags = ["-Dz", "-bbinary"]
+    if config.arch.big_endian:
+        objdump_flags.append("-EB")
+    else:
+        objdump_flags.append("-EL")
```
simonlindholm, line 1318:
```suggestion
    objdump_flags = ["-Dz", "-bbinary"] + ["-EB" if config.arch.big_endian else "-EL"]
```

simonlindholm (approved):

PR #61, opened 2021-09-19
Always parse line/source info from objdump output

zbanks:
Rename `config.source` --> `config.show_source`.

Even if `config.show_source` or `config.show_line_numbers` is disabled, still parse the line/source information in the objdump output. This makes the information available through the JSON interface, without showing up in the formatting.

The thing that makes me the most nervous about this PR is the refactor around lines 1777 or so -- using the same regex regardless of `config.source_old_binutils`. I'd consider this change to be a lot less conservative, but I can't figure out a situation that breaks anything.

So far I've tested thoroughly with `mips`, and I want to test some `aarch64` before merging.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1774,7 +1774,7 @@ def process(dump: str, config: Config) -> List[Line]:
             )
             break
 
-        if row and not re.match(r"^ +[0-9a-f]+:\t", row):
+        if not re.match(r"^ +[0-9a-f]+:\t", row):
             # This regex is conservative, and assumes the file path does not contain "weird"
             # characters like colons, tabs, or angle brackets.
             if re.match(
--- a/diff.py
+++ b/diff.py
@@ -1748,7 +1748,10 @@ def process(dump: str, config: Config) -> List[Line]:
         row = lines[i]
         i += 1
 
-        if config.diff_obj and (">:" in row or not row):
+        if not row:
+            continue
+
+        if re.match(r"^[0-9a-f]+ <.*>:$", row):
             continue
 
         if row.startswith("DATAREF"):
--- a/diff.py
+++ b/diff.py
@@ -120,7 +120,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "-c",
         "--source",
-        dest="source",
+        dest="show_source",
         action="store_true",
         help="Show source code (if possible). Only works with -o or -e.",
     )
@@ -380,7 +380,6 @@ class Config:
     # Build/objdump options
     diff_obj: bool
     make: bool
-    source: bool
     source_old_binutils: bool
     inlines: bool
     max_function_size_lines: int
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -965,10 +965,11 @@ def serialize_data_references(references: List[Tuple[int, int, str]]) -> str:
 def maybe_get_objdump_source_flags(config: Config) -> List[str]:
     flags = []
 
-    if config.show_line_numbers or config.source:
+    if config.show_line_numbers or config.show_source or True:
```
leoetlino, line 968: Doesn't this make the condition always true and cause `--line-numbers` and `--source` to be always passed? If so, why not just remove the if altogether?

That said, unconditionally passing those source flags is *highly* undesirable for larger projects (e.g. BotW). As mentioned last time we discussed the `show_line_numbers` default, objdump + --source takes an unacceptable amount of time for interactive diffing of larger functions: 3s+ for a 30kB function, and 42s on an extreme case (150kB function).

I'd prefer to have a --always-parse-source flag instead of forcing options that cause a dramatic slowdown for larger projects... (a significant time cost that just goes to waste if the source info isn't even printed in the end).

diff.py
```diff
@@ -965,10 +965,11 @@ def serialize_data_references(references: List[Tuple[int, int, str]]) -> str:
...
         flags.append("--line-numbers")
 
-    if config.source:
+    config.source_old_binutils = True
```
leoetlino, line 971: Was this added for testing and supposed to be removed before committing? ;)

FYI, I hadn't tested source_old_binutils so far, but the lack of --source-comment seems to prevent asm-differ from distinguishing between empty source lines and empty objdump lines — which causes broken output with richer debug info that contains empty source lines. The latter show up as added lines in this branch.

diff.py
zbanks: Both changes in this fn were from testing, my mistake for including them (committed at the wrong time!)

diff.py
zbanks: Aside from the mistakes I had previously included, is the behavior different from the version on `main`? 

For me, on binutils version `2.26.1` ("old"), `objdump` skips printing empty source lines.

zbanks: I had been testing with IDO for `mips`, and GCC for `aarch64`, with objdumps version 2.26 & 2.37.

I was able to reproduce this by not using the `diff_obj` mode: there is an earlier check that stripped empty lines, but only if `config.diff_obj` was true. I refactored that to apply regardless of mode. 

This now mostly fixes the problem that exists on `main` for `--source-old-binutils`. One minor drawback is that now with this flag, completely empty source lines will be excluded from the source view in `diff.py`'s output. (This is only with `--source-old-binutils`; blank lines are included with `--source`, so this seems like a reasonable tradeoff)

leoetlino: Yeah that sounds reasonable since it's not like we can tell if the blank line is source code or not. (I wonder why there was a check for obj mode though 🤔)

simonlindholm (approved):

zbanks:
As far as I can tell, it should be good to know, but let me know @leoetlino if you run into any further issues!

PR #63, opened 2021-09-26
Annotate data pool symbols inline for ARM32

abaresk:
References data pool symbols inline for ARM32. Displays the symbol and the address (parenthesized). Here is an example snippet (see [full diff](https://gist.github.com/abaresk/33b611ac8fb35f7976dc43d35c86689b)):

```
b4a:    ldr     r2, [pc, #0x1c] ; =gBlockRecvBuffer (b68)
...                              
b68:    .word   gBlockRecvBuffer   
```

Created a `RelocationProcessor` class which each architecture implements. If an asm line matches the architecture's `re_reloc` regex, `reloc_processor.process()` gets called on that line. This makes it easier to trigger on the relocation sentinel.

Moved existing relocation functionality for non-ARM32 architectures into their own sub-classes.

Created a `PostProcessor` class, which is called at the end of `process()`. This is useful if you need access to all the output lines to perform any processing. For ARM32, objdump only outputs the addresses of data pool symbols. So in `PostProcessorARM32`, we dereference the symbol at the given data pool address. E.g. objdump output:

```
00000adc <Task_LinkContest_CommunicateTurnOrder>:
 ...
 b4a:   4a07            ldr     r2, [pc, #28]   ; (b68 <Task_LinkContest_CommunicateTurnOrder+0x8c>)
 b4c:   1889            adds    r1, r1, r2
 ...
 b68:   00000000        .word   0x00000000
                        b68: R_ARM_ABS32        gBlockRecvBuffer
``` 

TESTS = [ARM32](https://gist.github.com/abaresk/33b611ac8fb35f7976dc43d35c86689b) (new functionality), [PPC](https://gist.github.com/abaresk/df4f30ebc693c035040b5c7d377eda79) (unchanged), [MIPS](https://gist.github.com/abaresk/7a0cdb31d4e73b64f689107595b108d6) (unchanged), AARCH64 doesn't process relocations

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1303,6 +1303,126 @@ def dump_binary(
         (objdump_flags + flags2, project.myimg, None),
     )
 
+RELOC_PLACEHOLDER = "RELOCATION_PLACEHOLDER-OFFSET_{}"
+RELOC_PLACEHOLDER_PATTERN = re.compile(r"RELOCATION_PLACEHOLDER-OFFSET_([a-zA-z0-9]+)")
+
+class RelocationProcessor:
+    def __init__(self, config: Config) -> None:
+        self.config = config
+
+    # The base class is a no-op.
+    def process(self, curLine: str, lines: List["Line"]) -> Optional[str]:
+        return curLine
+
+
+class RelocationProcessorMIPS(RelocationProcessor):
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
+
+    def process(self, curLine: str, lines: List["Line"]) -> Optional[str]:
+        # Remove previous line in order to replace it with reloc-processed line
+        prevLineIdx = len(lines) - 1
+        prevLine = lines.pop(prevLineIdx)
+        return self._process_mips_reloc(curLine, prevLine, self.config.arch)
+
+    def _process_mips_reloc(self, row: str, prev: str, arch: "ArchSettings") -> str:
+        before, imm, after = parse_relocated_line(prev)
+        repl = row.split()[-1]
+        if imm != "0":
+            # MIPS uses relocations with addends embedded in the code as immediates.
+            # If there is an immediate, show it as part of the relocation. Ideally
+            # we'd show this addend in both %lo/%hi, but annoyingly objdump's output
+            # doesn't include enough information to pair up %lo's and %hi's...
+            # TODO: handle unambiguous cases where all addends for a symbol are the
+            # same, or show "+???".
+            mnemonic = prev.split()[0]
+            if (
+                mnemonic in arch.instructions_with_address_immediates
+                and not imm.startswith("0x")
+            ):
+                imm = "0x" + imm
+            repl += "+" + imm if int(imm, 0) > 0 else imm
+        if "R_MIPS_LO16" in row:
+            repl = f"%lo({repl})"
+        elif "R_MIPS_HI16" in row:
+            # Ideally we'd pair up R_MIPS_LO16 and R_MIPS_HI16 to generate a
+            # correct addend for each, but objdump doesn't give us the order of
+            # the relocations, so we can't find the right LO16. :(
+            repl = f"%hi({repl})"
+        elif "R_MIPS_26" in row:
+            # Function calls
+            pass
+        elif "R_MIPS_PC16" in row:
+            # Branch to glabel. This gives confusing output, but there's not much
+            # we can do here.
+            pass
+        else:
+            assert False, f"unknown relocation type '{row}' for line '{prev}'"
+        return before + repl + after
+
+
+class RelocationProcessorPPC(RelocationProcessor):
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
+
+    def process(self, curLine: str, lines: List["Line"]) -> Optional[str]:
+        # Remove previous line in order to replace it with reloc-processed line
+        prevLineIdx = len(lines) - 1
+        prevLine = lines.pop(prevLineIdx)
+        return self._process_ppc_reloc(self, curLine, prevLine)
+
+    def _process_ppc_reloc(self, row: str, prev: str) -> str:
+        assert any(
+            r in row for r in ["R_PPC_REL24", "R_PPC_ADDR16", "R_PPC_EMB_SDA21"]
+        ), f"unknown relocation type '{row}' for line '{prev}'"
+        before, imm, after = parse_relocated_line(prev)
+        repl = row.split()[-1]
+        if "R_PPC_REL24" in row:
+            # function calls
+            pass
+        elif "R_PPC_ADDR16_HI" in row:
+            # absolute hi of addr
+            repl = f"{repl}@h"
+        elif "R_PPC_ADDR16_HA" in row:
+            # adjusted hi of addr
+            repl = f"{repl}@ha"
+        elif "R_PPC_ADDR16_LO" in row:
+            # lo of addr
+            repl = f"{repl}@l"
+        elif "R_PPC_ADDR16" in row:
+            # 16-bit absolute addr
+            if "+0x7" in repl:
+                # remove the very large addends as they are an artifact of (label-_SDA(2)_BASE_)
+                # computations and are unimportant in a diff setting.
+                if int(repl.split("+")[1], 16) > 0x70000000:
+                    repl = repl.split("+")[0]
+        elif "R_PPC_EMB_SDA21" in row:
+            # small data area
+            pass
+        return before + repl + after
+
+
+class RelocationProcessorARM32(RelocationProcessor):
+    def __init__(self, config: Config) -> None:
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1323,6 +1323,126 @@ def dump_binary(
...
 
+RELOC_PLACEHOLDER = "RELOCATION_PLACEHOLDER-OFFSET_{}"
+RELOC_PLACEHOLDER_PATTERN = re.compile(r"RELOCATION_PLACEHOLDER-OFFSET_([a-zA-z0-9]+)")
+
+class RelocationProcessor:
+    def __init__(self, config: Config) -> None:
+        self.config = config
+
+    # The base class is a no-op.
+    def process(self, curLine: str, lines: List["Line"]) -> Optional[str]:
```
simonlindholm, line 1334: curLine -> cur_line, similar elsewhere

diff.py
```diff
@@ -1406,7 +1526,6 @@ def _normalize_bl(self, mnemonic: str, row: str) -> str:
         row, _ = split_off_address(row)
         return row + "<ignore>"
 
-
```
simonlindholm, line 1409: (this newline will get added back the next time someone runs `black`)

diff.py
```diff
@@ -1824,21 +1884,9 @@ def process(dump: str, config: Config) -> List[Line]:
...
-            elif "R_PPC_" in reloc_row:
-                original = process_ppc_reloc(reloc_row, original)
-            elif "R_ARM_" in reloc_row:
-                original = process_arm_reloc(reloc_row, original, arch)
-            else:
-                break
-            i += 1
+        if re.search(arch.re_reloc, row):
+            processed_row = reloc_processor.process(row, output)
+            original = processed_row if processed_row else row
```
simonlindholm, line 1889: I'm not too happy about making this a post-processing step that fixes up the previous row. We had it like that before, and it resulted in confusion and bugs. For instance you have the following bugs with this version of the code:

- `data_refs` will apply twice: one to the real line, once to the relocation
- `mnemonic` will be set wrongly (because this code runs too late) (but if you move it up much higher the hexify_int step will happen after relocation processing which I'm not sure if it has side effects)
- `comment`s get stripped I think
- `lui $v0, %hi(var)`, `jr $ra`, `addiu $v0, $v0, %lo(var)` won't process relocations for the last line with MIPS and --stop-jr-ra.

Perhaps more importantly the current way *feels* more right to me: the relocations are processed together with the current instruction as part of parsing it.

Using a separate class for processing relocations is a nice idea. I think you can keep that but change it to use the previous system? Would be fine to merge with DifferenceNormalizer too I think. ARM32_LOAD_POOL_PATTERN will need special treatment but that's fine: it isn't a relocation, it's something else.

diff.py
```diff
@@ -2593,6 +2670,7 @@ class Display:
     def __init__(self, basedump: str, mydump: str, config: Config) -> None:
         self.config = config
         self.base_lines = process(basedump, config)
+        post_process(self.base_lines)
```
simonlindholm, line 2673: move this call to the end of `process`; it's not great to require the caller to call both of process and post_process for correctness

diff.py
```diff
@@ -1899,6 +1947,35 @@ def process(dump: str, config: Config) -> List[Line]:
...
+        line_number = re.search(RELOC_PLACEHOLDER_PATTERN, reloc).group(1)
+        line_original = get_line(lines, int(line_number, 16)).original
+        value = line_original.split()[1]
+
+        # Replace relocation placeholder with value
+        replaced = re.sub(RELOC_PLACEHOLDER_PATTERN, 
+                          f"={value} ({line_number})", line.original)
+        line.original = replaced
+
+def get_line(lines: List[Line], line_number: int) -> Optional[Line]:
```
simonlindholm, line 1972: inline this within post_process_relocations, and make it use a lookup table so it's not O(n^2)

diff.py
```diff
@@ -1899,6 +1947,35 @@ def process(dump: str, config: Config) -> List[Line]:
     return output
 
 
+# Post-process the asm lines in-place. Includes:
+#  - De-referencing relocations
+def post_process(lines: List[Line]):
+    post_process_relocations(lines)
+
+def post_process_relocations(lines: List[Line]):
```
simonlindholm, line 1955: This seems specific to ARM32, should it be part of the ARM32 config? (or DifferenceNormalizer)

diff.py
abaresk: Done.

diff.py
abaresk: Reverted and re-ran with `black`

diff.py
abaresk: Done.

diff.py
abaresk: Done.

diff.py
abaresk: I kept the relocation classes and moved the logic marking data pool annotations into DifferenceNormalizer. I added a PostProcessor class -- for ARM32, it adds symbols in the final annotation by checking all the output lines and dereferencing addresses.

diff.py
abaresk: I created a PostProcessor class and moved this logic under PostProcessorARM32. Please let me know if you have any concerns with this.

diff.py
```diff
@@ -1835,21 +1932,11 @@ def process(dump: str, config: Config) -> List[Line]:
         # transforming 'row' into a coarser version that ignores registers and
         # immediates.
         original = row
-
-        while i < len(lines):
+        if i < len(lines):
```
simonlindholm, line 1935: this does need to be a while -- there can be multiple relocations that apply to the same line. In practice the case I know of is a silly one where all but one relocation are "no-op" relocations, so the duplicate processing is harmless

diff.py
```diff
@@ -1422,12 +1582,15 @@ class ArchSettings:
...
     re_imm: Pattern[str]
+    re_reloc: Pattern[str]
     branch_instructions: Set[str]
     instructions_with_address_immediates: Set[str]
     forbidden: Set[str] = field(default_factory=lambda: set(string.ascii_letters + "_"))
     arch_flags: List[str] = field(default_factory=list)
     branch_likely_instructions: Set[str] = field(default_factory=set)
     difference_normalizer: Type[DifferenceNormalizer] = DifferenceNormalizer
+    reloc_proc: Type[RelocationProcessor] = RelocationProcessor
+    post_proc: Type[PostProcessor] = PostProcessor
```
simonlindholm, line 1593: Can you fold all of these three processor classes into a single class? They fill more or less the same purpose of acting as arch hooks for the asm parsing process.

(Alternatively reoc_proc and post_proc could be made into functions, since the classes are stateless)

diff.py
```diff
@@ -1330,6 +1330,114 @@ def dump_binary(
...
+        self.config = config
+
+    # The base class is a no-op.
+    def process(self, row: str, prev: str, arch: "ArchSettings") -> str:
+        return prev
+
+
+class RelocationProcessorMIPS(RelocationProcessor):
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
```
simonlindholm, line 1344: can just inherit the constructor

diff.py
```diff
@@ -1330,6 +1330,114 @@ def dump_binary(
     )
 
 
+class RelocationProcessor:
+    def __init__(self, config: Config) -> None:
```
simonlindholm, line 1334: better pass arch as an argument here than in process

diff.py
```diff
@@ -1412,6 +1521,57 @@ def _normalize_bl(self, mnemonic: str, row: str) -> str:
...
+
+            # Replace relocation placeholder with value
+            replaced = re.sub(
+                DATA_POOL_PLACEHOLDER_PATTERN,
+                f"={value} ({line_number})",
+                line.normalized_original,
+            )
+            line.original = replaced
+
+    def extract_lines_by_line_number_(self, lines: List["Line"]) -> Dict[int, "Line"]:
```
simonlindholm, line 1569: not worth a separate function for this when it's 3 lines of code and used from only one place; inline it

diff.py
abaresk: Done, reverted.

diff.py
abaresk: Done, folded them into a single class.

diff.py
abaresk: Done (in the new `AsmProcessor` class).

diff.py
abaresk: I removed the argument from process() altogether, as it can be derived from `config`.

diff.py
abaresk: Done.

simonlindholm (approved):
Looks good. Left one review comment which I'm very unsure about because I haven't tested, and not it's important, just an output quality thing. I'll merge this regardless because it's stalled for so long and let you do a follow-up PR if you care.

PR #65, opened 2021-10-17
Remove extra trailing nops from either side and add max_score field

### LINKED ISSUE #60: Scoring fails to ignore extra nops at the end of the code on one side
https://decomp.me/scratch/3ZUPF

Technically there are deletions here, but imo it's safe to ignore extra nops for scoring purposes

ethteck:
Fixes #60

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -640,6 +640,7 @@ class Text:
 class TableMetadata:
     headers: Tuple[Text, ...]
     current_score: int
+    max_score: int
     previous_score: Optional[int]
 
 
@@ -832,6 +833,7 @@ class JsonFormatter(Formatter):
             for h, name in zip(meta.headers, ("base", "current", "previous"))
         }
         output["current_score"] = meta.current_score
+        output["max_score"] = meta.max_score
         if meta.previous_score is not None:
             output["previous_score"] = meta.previous_score
         output_rows: List[Dict[str, Any]] = []
@@ -2124,7 +2126,11 @@ class OutputLine:
 class Diff:
     lines: List[OutputLine]
     score: int
+    max_score: int
 
+
+def is_nops(lines: List[Line]) -> bool:
+    return all(line.mnemonic == "nop" for line in lines)
 
 def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
     if config.show_source:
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -2153,8 +2159,14 @@ def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
                     btset.add(bt)
                     sc(str(bt))
 
+    if len(lines1) > len(lines2) and is_nops(lines1[len(lines2):]):
+        lines1 = lines1[:len(lines2)]
+    elif len(lines1) < len(lines2) and is_nops(lines2[len(lines1):]):
+        lines2 = lines2[:len(lines1)]
```
simonlindholm, line 2165: This only trims nops in the zero-score case. We should make it work in other cases too. E.g. based on based on the last chunk of the computed diff, or trimming lines1 and lines2 until they contain equally many trailing nops

diff.py
```diff
@@ -1429,7 +1429,10 @@ class ArchSettings:
     branch_likely_instructions: Set[str] = field(default_factory=set)
     difference_normalizer: Type[DifferenceNormalizer] = DifferenceNormalizer
     big_endian: Optional[bool] = True
+    delay_slot_instructions: Set[str] = field(default_factory=set)
 
+    def has_delay_slot(self, mnemonic: str) -> bool:
```
simonlindholm, line 1434: I don't think this member function is helpful; just inline it at the callsite

diff.py
```diff
@@ -2129,8 +2133,12 @@ class Diff:
     max_score: int
 
 
-def is_nops(lines: List[Line]) -> bool:
-    return all(line.mnemonic == "nop" for line in lines)
+def get_trim_point(lines: List[Line], arch_settings: ArchSettings) -> bool:
```
simonlindholm, line 2136: this won't type-check! (run `mypy`) But also, I think `lines = trim_nops(lines, arch)` would be a clearer API, "trim point" is a bit vague. arch_settings -> arch for consistency

diff.py
```diff
@@ -2129,8 +2133,12 @@ class Diff:
...
+    for index, line in enumerate(reversed(lines)):
+        rindex = len(lines) - index - 1
```
simonlindholm, line 2138:
```suggestion
    for rindex, line in reversed(enumerate(lines)):
```
(and rename rindex -> index)

diff.py
```diff
@@ -2129,8 +2133,12 @@ class Diff:
...
+        if line.mnemonic != "nop" or arch_settings.has_delay_slot(lines[rindex - 1].mnemonic):
```
simonlindholm, line 2139: check that rindex - 1 isn't out of bounds before dereferencing

diff.py
```diff
@@ -2131,11 +2131,11 @@ class Diff:
 
 
 def trim_nops(lines: List[Line], arch: ArchSettings) -> List[Line]:
-    for index, line in reversed(list(enumerate(lines))):
-        rindex = len(lines) - index - 1
-        if line.mnemonic != "nop" or (rindex > 0 and lines[rindex - 1].mnemonic in arch.delay_slot_instructions):
-            return lines[:rindex + 1]
-    return []
+    while lines:
+        if lines[-1].mnemonic != "nop" or (len(lines) > 1 and lines[-2].mnemonic in arch.delay_slot_instructions):
```
simonlindholm, line 2135: fold this check into the while condition

diff.py
```diff
@@ -2131,11 +2131,11 @@ class Diff:
 
 
 def trim_nops(lines: List[Line], arch: ArchSettings) -> List[Line]:
```
simonlindholm, line 2133: `lines = lines[:]` to avoid mutating the input in-place

PR #67, opened 2021-11-27
added an option to restrict diff to a section, and little endian ARM

gamestabled:
The -j option (borrowing the letter name from objdump) can be used to specify a section to diff. Useful for certain project setups

"armel" added as an arch option for little endian ARM32

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1270,7 +1270,12 @@ def dump_elf(
         f"--disassemble={diff_elf_symbol}",
     ]
 
-    objdump_flags = ["-drz", "-j", ".text"]
+    if config.diff_section is not None:
+        section = config.diff_section
+    else:
+        section = ".text"
+
+    objdump_flags = ["-drz", "-j", section]
     return (
         project.myimg,
         (objdump_flags + flags1, project.baseimg, None),
@@ -1306,7 +1311,12 @@ def dump_objfile(
     if not os.path.isfile(refobjfile):
         fail(f'Please ensure an OK .o file exists at "{refobjfile}".')
 
-    objdump_flags = ["-drz", "-j", ".text"]
+    if config.diff_section is not None:
+        section = config.diff_section
+    else:
+        section = ".text"
+
+    objdump_flags = ["-drz", "-j", section]
     return (
         objfile,
         (objdump_flags, refobjfile, start),
@@ -1330,7 +1340,7 @@ def dump_binary(
         end_addr = eval_int(end, "End address must be an integer expression.")
     else:
         end_addr = start_addr + config.max_function_size_bytes
-    objdump_flags = ["-Dz"] + ["-belf32-littlearm" if config.arch.name == "armel" else "-bbinary"] + ["-EB" if config.arch.big_endian else "-EL"]
+    objdump_flags = ["-Dz", "-bbinary"] + ["-EB" if config.arch.big_endian else "-EL"]
     flags1 = [
         f"--start-address={start_addr + config.base_shift}",
         f"--stop-address={end_addr + config.base_shift}",
--- a/diff.py
+++ b/diff.py
@@ -133,6 +133,13 @@ if __name__ == "__main__":
         Implies --source.""",
     )
     parser.add_argument(
+        "-j",
+        "--section",
+        dest="diff_section",
+        metavar="SECTION",
+        help="Diff restricted to a given output section.",
+    )
+    parser.add_argument(
         "-L",
         "--line-numbers",
         dest="show_line_numbers",
@@ -381,6 +388,7 @@ class Config:
     diff_obj: bool
     make: bool
     source_old_binutils: bool
+    diff_section: str
     inlines: bool
     max_function_size_lines: int
     max_function_size_bytes: int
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1318,12 +1330,14 @@ def dump_binary(
         end_addr = eval_int(end, "End address must be an integer expression.")
     else:
         end_addr = start_addr + config.max_function_size_bytes
-    objdump_flags = ["-Dz", "-bbinary"] + ["-EB" if config.arch.big_endian else "-EL"]
+    objdump_flags = ["-Dz"] + ["-belf32-littlearm" if config.arch.name == "armel" else "-bbinary"] + ["-EB" if config.arch.big_endian else "-EL"]
```
simonlindholm, line 1333: This is somewhat of an abuse of `dump_binary`, making it not dump a binary range at all but rather parse an ELF file... I'm confused as to how this works when you're also passing in --start-address= and --stop-address=. Can you explain your workflow? Any chance you can get your project set up to use dump_objfile or dump_elf instead, which should be more user-friendly? e.g. dump_objfile/-o accepts a function name as argument which is nicer than a range of bytes.

simonlindholm (commented):
dump_elf already passes in `-j .text`, you'll need to change that to `"-j" + config.diff_section`

Are there multiple sections of interest, or should this be a project flag instead of a command line one? If it's not just boilerplate it seems like it would be nice to auto-detect which section to look at rather than have to specify it.

diff.py
gamestabled: For our project, we need to control the global ordering of functions at the final link stage, by essentially hard-coding the ordering. The best way for us to do this is to specify `--split_sections` to the compiler so that it generates an elf section for each function, then use a linkerscript to order those sections into the global `.text` section. So, each function `foo` gets placed in an elf section called `i.foo`. Then, "non-matchings" use the same section names, so that the linker script doesnt need to be edited to replace a non-matching with decompiled code.

Anyway, I was able to tweak our project setup to make it work with dump_elf, or maybe dump_obj with a bit more work, which is better.

diff.py
```diff
@@ -1258,7 +1270,12 @@ def dump_elf(
         f"--disassemble={diff_elf_symbol}",
     ]
 
-    objdump_flags = ["-drz", "-j", ".text"]
+    if config.diff_section is not None:
+        section = config.diff_section
+    else:
+        section = ".text"
```
simonlindholm, line 1276:
```suggestion
    section = config.diff_section or ".text"
```

diff.py
```diff
@@ -984,6 +993,9 @@ def maybe_get_objdump_source_flags(config: Config) -> List[str]:
         if config.inlines:
             flags.append("--inlines")
 
+    if config.diff_section is not None:
+        flags.append("-j" + config.diff_section)
```
simonlindholm, line 997: this shouldn't be needed, both callers of `maybe_get_objdump_source_flags` already pass this -j flag

diff.py
```diff
@@ -1324,6 +1346,8 @@ def dump_binary(
         f"--stop-address={end_addr + config.base_shift}",
     ]
     flags2 = [f"--start-address={start_addr}", f"--stop-address={end_addr}"]
+    if config.diff_section is not None:
+        flags2 = flags2 + [f"-j" + config.diff_section]
```
simonlindholm, line 1350: doesn't make sense for dump_binary

simonlindholm (commented):
There's also some references to .text in `parse_elf_data_references` and `search_map_file`

gamestabled:
I dont have any projects set up to be able to test the changes to search_map_file for the mw map format, so i would really appreciate if somebody could look at that closely 🙏

simonlindholm:
> `address feedback`

I think you accidentally committed the wrong file!

diff.py
```diff
@@ -132,6 +132,13 @@ def complete_symbol(
         help="""Tweak --source handling to make it work with binutils < 2.33.
         Implies --source.""",
     )
+    parser.add_argument(
+        "-j",
+        "--section",
+        dest="diff_section",
+        metavar="SECTION",
+        help="Diff restricted to a given output section.",
```
simonlindholm, line 140: there's a lot of `diff_section or ".text"`, can we do `default=".text"` here instead?

diff.py
```diff
@@ -1080,13 +1090,14 @@ def search_map_file(
         if len(cands) == 1:
             return cands[0]
     elif project.map_format == "mw":
+        section_patten = config.diff_section or r"\.(?:init|text)"
```
simonlindholm, line 1093: nit: section_pattern

I wonder what the .init thing here is about... seems like it doesn't work given our .text restriction. I'd be inclined to remove it and see if anyone yells

use `re.escape(config.diff_section)` to avoid the `.` being interpreted as a regex part

diff.py
```diff
@@ -1186,7 +1197,8 @@ class Section:
     assert len(symtab_sections) == 1
     symtab = sections[symtab_sections[0]]
 
-    text_sections = [i for i in range(e_shnum) if sec_names[i] == b".text" and sections[i].sh_size != 0]
+    section_name = config.diff_section or b".text"
```
simonlindholm, line 1200: `diff_section.encode("utf-8")`

PR #70, opened 2022-01-11
Fix data pool lookup errors

abaresk:
Fixes dictionary key lookup errors that occur when trying to dereference a data symbol (see example [scratch](https://decomp.me/scratch/XRuTB)). The root cause of the issue was [these lines](https://github.com/simonlindholm/asm-differ/blob/89df9d720d996267095edab0f17dc7026744e83e/diff.py#L1894-L1895), which tried to hexify the data symbol address even though objdump already dumps it as a hex value. As a result, it sometimes would try to lookup the symbol at an address that doesn't exist and crash.

Fixed this by extracting the data symbol address while parsing the asm lines and storing it in the `Line` data struct. Also removed the code for temporarily annotating a `DATA_POOL_PLACEHOLDER`, as this data is now directly available in the `Line`.

See scratch with fix applied:
Screen Shot 2022-01-11 at 9 58 38 AM

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1333,11 +1333,6 @@ def dump_binary(
         (objdump_flags + flags2, project.myimg, None),
     )
 
-DATA_POOL_PLACEHOLDER = "DATA_POOL_PLACEHOLDER-OFFSET_{}"
-DATA_POOL_PLACEHOLDER_PATTERN = re.compile(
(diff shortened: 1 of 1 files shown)


simonlindholm (approved):
Much cleaner!

PR #71, opened 2022-01-15
Fix the AArch64 register regex for FP regs followed by data width

leoetlino:
FP registers may be followed by data width and number of elements,
e.g. V0.4S (instead of just V0). For simplicity, we don't actually
check if the combination of data width / number of elements makes sense
or list all the combinations in the regex -- instead, the regex simply
looks for one or two digits followed by a valid letter.

Before:

before

After:

after


simonlindholm (approved):
(could simplify the regex by parenthesizing `([12]?[0-9]|3[01])` or loosening it to `[1-3]?[0-9]`)

PR #74, opened 2022-01-29
thumb support

notyourav:


diff.py
```diff
@@ -1873,7 +1877,7 @@ def process(dump: str, config: Config) -> List[Line]:
         # the symbol.
         data_pool_addr = None
         pool_match = re.search(ARM32_LOAD_POOL_PATTERN, row)
-        if pool_match:
+        if pool_match and arch.name != "thumb":
```
notyourav, line 1880: this breaks on armv4t, not sure if theres a better way to address it

PR #76, opened 2022-02-04
Implemented mips call16 relocs, ignored mips jalr relocs as they are just a linker optimization hint

Mr-Wiseguy:


ethteck:
it gud plz merg

PR #77, opened 2022-02-15
Allow diffing specific object files

jdflyer:
When all the functions in a project aren't in a single map file it can be useful to define the object file used when using -o instead of doing something like searching every map file in the project. This pull requests implements an extra argument to diff.py that allows the user to give the path of an object file to be diffed.


simonlindholm (approved):
This is great! I've needed this a bunch of times for local testing but always been lazy and hacked around it.

Having a flag for which .o to diff against could potentially be useful too.

PR #78, opened 2022-02-21
Ignore non-.rodata sections in diff

mkst:
Think that this is what you were suggesting I do when I added the `.mwcats.text` check...


simonlindholm (approved):
Can you rename parse_elf_data_references -> parse_elf_rodata_references at the same time? (and maybe similarly in whatever uses that function)

PR #79, opened 2022-02-22
Add support for MIPS ABI register names

EllipticEllipsis:
- Added `objdump_flags` to enable passing extra flags to objdump. 
- In particular this enables a project to set up a config for passing `-M reg-names=32` or similar, allowing for use of the ABI names for float registers (`fv0` and so on) instead of the numeric versions.
- Updated the regex to include these new names.
- Also added a note on what the regex matches (I noted that `gp` was not included, although maybe that's just because nothing uses it?).

Also followed the readme and ran `black`, which reformatted a bit more than I expected!

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -373,6 +373,7 @@ except ModuleNotFoundError as e:
 class ProjectSettings:
     arch_str: str
     objdump_executable: str
+    objdump_flags: List[str]
     build_command: List[str]
     map_format: str
     mw_build_dir: str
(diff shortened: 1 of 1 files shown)


simonlindholm:
> (I noted that gp was not included, although maybe that's just because nothing uses it?).

For sp the reason is that an sp vs other reg diff shouldn't show up as `r` because it's a semantic difference. For gp I don't remember if it's unintentional or intentional with similar reasoning to sp. It seems okay to leave out of the regex, in any case.

PR #81, opened 2022-03-12
Scoring fixes

leoetlino:
This PR fixes latent bugs in the scoring and branch target normalisation logic that have been exposed by a recent change (b8907ddfc7?)

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2044,7 +2044,7 @@ def split_off_address(line: str) -> Tuple[str, str]:
         parts = line.split(None, 1)
         if len(parts) < 2:
             parts.append("")
-    off = len(line) - len(parts[-1])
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -2045,7 +2045,7 @@ def split_off_address(line: str) -> Tuple[str, str]:
         if len(parts) < 2:
             parts.append("")
     off = len(line) - len(parts[-1])
-    return line[:off], line[off:]
+    return line[:off], line[off:].strip()
```
simonlindholm, line 2048: Can we do `off = len(line) - len(parts[-1].strip())` instead, to maintain the property that concatenating the resulting pair gives the original string back?

leoetlino: Done

PR #82, opened 2022-03-27
Always score --max-size lines when specified

ethteck:


ethteck:
per discord discussion, this fix is not really ideal. maybe we can solve the underlying problem in a better way later on

PR #83, opened 2022-05-01
Fix function look-up for certain map files

Xeeynamo:
asm-differ always assumes that a function name in a `.map` file is in the format `ADDRESS FUNCTION`. But in my case some functions has the format `ADDRESS FUNCTION = ADDRESS`, which were not previously detected. This commit fixes that bug.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1095,7 +1095,7 @@ def search_map_file(
                     ram = int(tokens[1], 0)
                     rom = int(tokens[5], 0)
                     ram_to_rom = rom - ram
-                if line.endswith(" " + fn_name):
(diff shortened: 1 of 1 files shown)


simonlindholm:
This will break when searching for a function names that are prefixes of others. Can we check `line.endswith(" " + x) or " " + x + " " in line` or something like that?

Xeeynamo:
I have a better proposal. Would `line.endswith(" " + x) or f" {fn_name} = 0x"` work? It will cover both cases and it is more strict.

simonlindholm:
Sure!

PR #84, opened 2022-05-28
fix ppc sda21 relocations

snuffysasa:
and fix wild card matches for ppc symbols that should match to everything

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1474,8 +1474,13 @@ class AsmProcessorPPC(AsmProcessor):
                 if int(repl.split("+")[1], 16) > 0x70000000:
                     repl = repl.split("+")[0]
         elif "R_PPC_EMB_SDA21" in row:
-            # small data area
-            pass
+            # With sda21 relocs, the linker transforms `r0` into `r2`/`r13`, and
+            # we may encounter this in either pre-transformed or post-transformed
+            # versions depending on if the .o file comes from compiler output or
+            # from disassembly. Normalize, to make sure both forms are treated as
+            # equivalent.
+            after = after.replace("(r2)", "(0)")
+            after = after.replace("(r13)", "(0)")
         return before + repl + after
 
 
@@ -2051,8 +2056,10 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
 
 
 def imm_matches_everything(row: str, arch: ArchSettings) -> bool:
-    # (this should probably be arch-specific)
-    return "(." in row
+    if arch.name == "ppc":
+        return any(str in row for str in ["...data", "@"])
+    else:
+        return "(." in row
 
 
 def split_off_address(line: str) -> Tuple[str, str]:
@@ -2189,6 +2196,8 @@ def score_diff_lines(
             newfields = newfields[:-1]
             oldfields = oldfields[:-1]
         for nf, of in zip(newfields, oldfields):
+            if imm_matches_everything(nf, config.arch):
+                continue
             if nf != of:
                 score += config.penalty_regalloc
         # Penalize any extra fields
@@ -2417,6 +2426,11 @@ def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
                         out1, out2 = format_fields(arch.re_reg, out1, out2, sc1, sc2)
                         line_color1 = line_color2 = sym_color = BasicFormat.REGISTER
                         line_prefix = "r"
+
+                if imm_matches_everything(branchless2, arch):
+                    # ignore differences due to %lo(.rodata + ...) vs symbol
+                    out1 = out1.reformat(BasicFormat.NONE)
+                    out2 = out2.reformat(BasicFormat.NONE)
 
                 if same_target:
                     address_imm_fmt = BasicFormat.NONE


diff.py
```diff
@@ -1857,6 +1865,7 @@ class Line:
     original: str
     normalized_original: str
     scorable_line: str
+    relocation_occured: bool
```
simonlindholm, line 1868: occurred. or `has_symbol`; may be clearer. or maybe even make this a `symbol: Optional[str]`? then `imm_matches_everything` could test for it without needing to care about arch-specific syntax for relocations

diff.py
```diff
@@ -2051,8 +2064,10 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
 
 
 def imm_matches_everything(row: str, arch: ArchSettings) -> bool:
-    # (this should probably be arch-specific)
-    return "(." in row
+    if arch.name == "ppc":
+        return any(str in row for str in ["...data", "@"])
```
simonlindholm, line 2068: I'm worried `@` may occur in the middle of symbol names

diff.py
```diff
@@ -2163,8 +2178,10 @@ def lo_hi_match(old: str, new: str) -> bool:
         new_inner = new[new_idx + 4 : -1]
         return old_inner.startswith(".") or new_inner.startswith(".")
 
-    def diff_sameline(old: str, new: str) -> None:
+    def diff_sameline(old_line_obj: Line, new_line_obj: Line) -> None:
```
simonlindholm, line 2181: no need for an _obj suffix

diff.py
```diff
@@ -2188,7 +2205,16 @@ def diff_sameline(old: str, new: str) -> None:
         if ignore_last_field:
             newfields = newfields[:-1]
             oldfields = oldfields[:-1]
+        else:
+            ### If the last field has a register suffix, i.e. "0x38(r7)"  we split that part out to make it a seperate field
```
simonlindholm, line 2209:
```suggestion
            # If the last field has a register suffix, i.e. "0x38(r7)"  we split that part out to make it a separate field
```

diff.py
```diff
@@ -2425,10 +2451,21 @@ def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
...
+                if imm_matches_everything(line2_parts[i], config.arch):
+                    # And line1 (left side) had a relocation, then ignore this mismatch
+                    if line1.relocation_occured:
+                        continue
+                if line1_parts[i] != line2_parts[i]:
+                    line_prefix = "|"
+                    line_color1 = line_color2 = sym_color = BasicFormat.DIFF_CHANGE
+                    out1 = out1.reformat(line_color1)
+                    out2 = out2.reformat(line_color2)
+                    break
```
simonlindholm, line 2468: Why is this needed? This is supposed to be handled by the `line1.diff_row == line2.diff_row` case

diff.py
simonlindholm: also splitting on "(" seems dangerous for mips, which uses syntax like `%lo(sym)` for relocs; for address modes you thus get something like `%lo(sym)($reg)`. It would be nice if we could store symbols away in a separate field and not have to re-parse them here

diff.py
snuffysasa: I did some searching on this, valid symbol names for asm, and also C++ mangling, and haven't seen anything yet that suggests a  @ can appear in a symbol.

even if it did... it would just make that symbol match any other symbol,

diff.py
snuffysasa: the reason I have it is because this function now contains old as a string, new as a string, and then two additional old new vars that are the actual object...   is it okay to just have those called old_line and new_line?

then we will have old (str), new (str), old_line (Line) new_line (Line)

I'll put it to whatever you decide.

diff.py
snuffysasa: I could be wrong, but I looked though a bunch of examples of mips asm, and I don't think it causes any issues.
Actually, in some cases it may lead to slightly more accurate scoring, if there are multiple diffs in the line.

diff.py
snuffysasa: messaged you on discord about this, think I got it resolved.

diff.py
zbanks: MWCC will use `@` in the symbol names of thunks in C++, [here's an example from SMS](https://github.com/doldecomp/sms/blob/master/asm/Camera/sunmodel.s#L1007).

diff.py
snuffysasa: nevermind, this is still needed, its that darn r2/0 r13/0 swaps.  I figured out some other workarounds to make this work in the  `line1.diff_row == line2.diff_row` case, however, I think you will not like those workarounds as much as what is here.

diff.py
snuffysasa: hmm, shoot! I'll work on another option.

snuffysasa: okay my latest commit addresses the concern of @s appearing in symbol names.

diff.py
snuffysasa: I started a new branch that worked like this method.. it took a good amount of reworking, but then the problem still remains we want to tell the difference between any any symbol from the relocation, and a symbol that is a catch anything symbol. And therefore we still need the logic to check for that somewhere.   And then also, doing this makes the method / process diverge more from the permutor code, which would take more reworking since it's not setup with an object that is accessed in both places.

diff.py
simonlindholm: > MWCC will use @ in the symbol names of thunks in C++, [here's an example from SMS](https://github.com/doldecomp/sms/blob/master/asm/Camera/sunmodel.s#L1007).

That name also starts with a `@`, so middle of name vs start doesn't matter for it. (But perhaps the heuristic "starts with @ = matches anything" is wrong? or can we treat this symbol as matching anything?)

snuffysasa: > > MWCC will use @ in the symbol names of thunks in C++, [here's an example from SMS](https://github.com/doldecomp/sms/blob/master/asm/Camera/sunmodel.s#L1007).
> 
> That name also starts with a `@`, so middle of name vs start doesn't matter for it. (But perhaps the heuristic "starts with @ = matches anything" is wrong? or can we treat this symbol as matching anything?)

I think it would not be the end of the world if that symbol did match to everything(and by everything I mean any other symbol), however, In this PR I already have it fixed so that these symbols will not pass the imm_matches_everything test. 

So these symbols `"@32@__dt__9TSunModelFv"` will not be detected as a match with anything symbol

diff.py
```diff
@@ -1769,7 +1770,7 @@ class ArchSettings:
     name="ppc",
     re_int=re.compile(r"[0-9]+"),
     re_comment=re.compile(r"(<.*>|//.*$)"),
-    re_reg=re.compile(r"\$?\b([rf][0-9]+)\b"),
+    re_reg=re.compile(r"(\$?\b([rf][0-9]+)\b|\(.+\))"),
```
simonlindholm, line 1773: Parentheses should not be included in register names. Use lookbehinds/lookaheads if you need to. Also this should only need to capture `0` which should make the regex easier to understand.

Do we actually need to change this regex though? If we don't it will treat `0` vs real register as an instruction diff, which is imo okay because they are categorically different. That's similar to how `$sp` isn't in re_regs for MIPS.

diff.py
```diff
@@ -1961,7 +1963,29 @@ def process(dump: str, config: Config) -> List[Line]:
             row_parts = [part.lstrip() for part in row.split(" ", 1)]
         mnemonic = row_parts[0].strip()
         args = row_parts[1] if len(row_parts) >= 2 else ""
-        row = mnemonic + "\t" + args.replace("\t", "  ")
+
+        if arch.name == "ppc":
```
simonlindholm, line 1967: this logic should go on a callback on AsmProcessorPPC

diff.py
```diff
@@ -1961,7 +1963,29 @@ def process(dump: str, config: Config) -> List[Line]:
             row_parts = [part.lstrip() for part in row.split(" ", 1)]
         mnemonic = row_parts[0].strip()
         args = row_parts[1] if len(row_parts) >= 2 else ""
-        row = mnemonic + "\t" + args.replace("\t", "  ")
```
simonlindholm, line 1964: this should still happen, even for non-ppc

diff.py
```diff
@@ -1961,7 +1963,29 @@ def process(dump: str, config: Config) -> List[Line]:
...
+
+        if arch.name == "ppc":
+            if mnemonic == "li":
+                mnemonic = "addi"
+                args_parts = args.split(",")
+                args = args_parts[0] + ",0," + args_parts[1]
+            if mnemonic == "lis":
+                mnemonic = "addis"
+                args_parts = args.split(",")
+                args = args_parts[0] + ",0," + args_parts[1]
```
simonlindholm, line 1975: This needs a comment. Also, can we do it just if there's a successive R_PPC_EMB_SDA21?

diff.py
```diff
@@ -2050,9 +2078,21 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
...
+            return True
+
+        field = "".join(field.rsplit("@ha", 1))
+        field = "".join(field.rsplit("@l", 1))
+        field = "".join(field.rsplit("@sda21", 1))
+
+        return re.search(re.compile(r"\A@\d+\Z"), field) != None
+    else:
+        return (
+            re.search(re.compile(r"\A%(hi|lo)\(\.[a-z]+(\+0x([a-f]|\d)+)?\)"), field)
```
simonlindholm, line 2093: this should be done just for mips, and the regex pre-compiled instead of compiled once per call

`([a-f]|\d)` -> `[a-f0-9]`

(imo the existing `"(." in row` is also just fine to keep, along with a comment saying what it targets. It has the advantage of less complexity)

diff.py
```diff
@@ -2050,9 +2078,21 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
     return re.sub(arch.re_sprel, "addr(sp)", row)
 
 
-def imm_matches_everything(row: str, arch: ArchSettings) -> bool:
-    # (this should probably be arch-specific)
-    return "(." in row
+def imm_matches_everything(field: str, arch: ArchSettings) -> bool:
+    if arch.name == "ppc":
```
simonlindholm, line 2082: can you make this a method on AsmProcessorPPC instead?

diff.py
```diff
@@ -2050,9 +2078,21 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
...
+        if "...data" in field:
+            return True
+
+        field = "".join(field.rsplit("@ha", 1))
+        field = "".join(field.rsplit("@l", 1))
+        field = "".join(field.rsplit("@sda21", 1))
```
simonlindholm, line 2088: `field = field.replace("@ha", "").replace("@l", "").replace("@sda21", "")` is easier to read (and kinda ugly that it needs to be done, but understandable...)

if you want it to happen only at the very end of the string, rsplit is not sufficient for that

diff.py
```diff
@@ -2050,9 +2078,21 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
...
+
+        return re.search(re.compile(r"\A@\d+\Z"), field) != None
```
simonlindholm, line 2090: `\A` -> `^`, `\Z` -> `$` and `!= None` -> `is not None` per usual python coding style. `match` or `fullmatch` make more sense than `search` if you want to check that the full string matches something

diff.py
simonlindholm: `old_line`/`new_line` are good names imo; they match the `Line` type

diff.py
```diff
@@ -2188,7 +2230,16 @@ def diff_sameline(old: str, new: str) -> None:
         if ignore_last_field:
             newfields = newfields[:-1]
             oldfields = oldfields[:-1]
+        else:
+            ### If the last field has a register suffix, i.e. "0x38(r7)"  we split that part out to make it a separate field
```
simonlindholm, line 2234:
```suggestion
            # If the last field has a register suffix, i.e. "0x38(r7)"  we split that part out to make it a separate field
```

diff.py
snuffysasa: I don't think we need this regex changed anymore. I will change it back

diff.py
snuffysasa: this resulted in some more refactoring, note the function pre_process gets called for all archs, and the default behavior is 

>row = mnemonic + "\t" + args.replace("\t", "  ")

but extra steps for ppc

diff.py
snuffysasa: hmmmm, I guess you are right!

diff.py
snuffysasa: aha, I changed it away from "(." in row, because you asked me too. Atleast that is what I understood

diff.py
snuffysasa: you mean like a static method?
the processor class object is not available everywhere this is called

diff.py
snuffysasa: I did it with rsplit so it would only pull from the very end and only once,

I wasn't sure if those C++ mangles symbol names could have this pattern in the middle, but looks like they very easily could.

So right now this works and correctly does not catch C++ mangled ones, but could cause issues if I make this change. So I will hold off. unless you tell me to.

diff.py
snuffysasa: ahh didn't know about fullmatch function

snuffysasa: If I was to precompile this, where to put it? it's unique to PPC. should I make it global constant? or have a blank field in the other classes, and what to call this field? "re.compile(r"^@\d+$")"

diff.py
snuffysasa: also I found if I set it to 
>field.replace("@ha", "").replace("@l", "").replace("@sda21", "")

black reformats it and makes it multi line anyways

diff.py
simonlindholm: Since this happens regardless of arch, move it to after `pre_process`

diff.py
```diff
@@ -1959,9 +1995,16 @@ def process(dump: str, config: Config) -> List[Line]:
...
         mnemonic = row_parts[0].strip()
         args = row_parts[1] if len(row_parts) >= 2 else ""
-        row = mnemonic + "\t" + args.replace("\t", "  ")
+
+        if i < len(lines):
+            next_row = lines[i]
+        else:
+            next_row = None
+
+        row = processor.pre_process(mnemonic, args, next_row)
```
simonlindholm, line 2007: mnemonic and args need to be updated too, while `row` can be recomputed from those two. So make `pre_process` return Tuple[str, str]?

diff.py
```diff
@@ -1474,8 +1506,9 @@ def process_reloc(self, row: str, prev: str) -> str:
                 if int(repl.split("+")[1], 16) > 0x70000000:
                     repl = repl.split("+")[0]
         elif "R_PPC_EMB_SDA21" in row:
-            # small data area
-            pass
+            # sda21 relocations; r2/r13 --> 0 swaps are performed in an earlier processing step
```
simonlindholm, line 1509:
```suggestion
            # sda21 relocations; r2/r13 --> 0 swaps are performed in pre_process
```

diff.py
```diff
@@ -1447,6 +1450,35 @@ def process_reloc(self, row: str, prev: str) -> str:
...
+            )
+
+            # We want to convert li and lis with an sda21 reloc,
+            # because the r0 to r2/r13 transformation results in
+            # turning an li/lis into an addi/addis with r2/r13 arg
+            # our preprocessing normalizes all versions to addi with a 0 arg
+            if mnemonic == "li" or mnemonic == "lis":
+                mnemonic = mnemonic.replace(
+                    "li", "addi"
+                )  # this also turns lis into addis
```
simonlindholm, line 1476: no need for this comment

diff.py
```diff
@@ -1447,6 +1450,35 @@ def process_reloc(self, row: str, prev: str) -> str:
...
+                .replace("(r13)", "(0)")
+                .replace(",r2,", ",0,")
+                .replace(",r13,", ",0,")
+            )
+
+            # We want to convert li and lis with an sda21 reloc,
+            # because the r0 to r2/r13 transformation results in
+            # turning an li/lis into an addi/addis with r2/r13 arg
+            # our preprocessing normalizes all versions to addi with a 0 arg
+            if mnemonic == "li" or mnemonic == "lis":
```
simonlindholm, line 1473:
```suggestion
            if mnemonic in {"li", "lis"}:
```

diff.py
simonlindholm: Could do
```py
parts = field.rsplit("@", 1)
if len(parts) == 2 and parts[1] in {"l", "h", "ha", "sda21"}:
    field = parts[0]
```

diff.py
simonlindholm: I was thinking AsmProcessorPPC, but since that's not available I guess we can skip manually caching the compiled regex and hoping Python's builtin regex compile cache will handle it

diff.py
snuffysasa: sounds good, before it was setup where it needed to happen inside of pre_process, but after lots of refactoring and the change you suggested about moving li-->addi to only in sda relocs.

Okay, but to move this like to after pre_process, it means preprocess will now need to return a tuple of updated mnemonic and args.

diff.py
snuffysasa: > (watch out for accidentally mentioning github users when typing `@`s!)

lol!! Whoops

snuffysasa: Okay I took your latest code

diff.py
```diff
@@ -1959,9 +1995,16 @@ def process(dump: str, config: Config) -> List[Line]:
...
             row_parts = [part.lstrip() for part in row.split(" ", 1)]
+
         mnemonic = row_parts[0].strip()
         args = row_parts[1] if len(row_parts) >= 2 else ""
-        row = mnemonic + "\t" + args.replace("\t", "  ")
+
+        if i < len(lines):
+            next_row = lines[i]
+        else:
+            next_row = None
```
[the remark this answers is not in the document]
snuffysasa: okay using that

diff.py
snuffysasa: looks like there is another var called `row_parts`  and now that does not get updates.. but looks like its only used for things downstream that wouldn't be affected... little confusing though

diff.py
```diff
@@ -2050,9 +2096,18 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
     return re.sub(arch.re_sprel, "addr(sp)", row)
 
 
-def imm_matches_everything(row: str, arch: ArchSettings) -> bool:
-    # (this should probably be arch-specific)
-    return "(." in row
+def imm_matches_everything(field: str, arch: ArchSettings) -> bool:
```
simonlindholm, line 2099: Naming this `field` and including logic here that assumes it's run on an entire field is misleading, because this function is called with an entire instruction row (well, almost; the branch target part is stripped) by the existing caller.

This also means the below logic won't work for highlighting, only for scoring. You should be able to search for the `@\d+@(l|h|ha|sda21)\b` regex instead

Speaking of the existing caller: that uses `re_imm` which lists ha, h, l as the possible relocs. Needs to include sda21 too

diff.py
```diff
@@ -2050,9 +2096,18 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
...
+    if arch.name == "ppc":
+        if "...data" in field:
+            return True
+
+        parts = field.rsplit("@", 1)
+        if len(parts) == 2 and parts[1] in {"l", "h", "ha", "sda21"}:
+            field = parts[0]
+
+        return re.fullmatch(re.compile(r"^@\d+$"), field) is not None
```
simonlindholm, line 2108: re.compile() can be removed, it's only needed if you want to compile the regex beforehand for perf/cleanliness

diff.py
simonlindholm: oh, good catch. Those usages of `row_parts[1]` should be changed to `args`. And `args` should probably be strip()ped at the place where it's defined so those strip() usages can go away too

diff.py
```diff
@@ -2188,7 +2245,16 @@ def diff_sameline(old: str, new: str) -> None:
         if ignore_last_field:
             newfields = newfields[:-1]
             oldfields = oldfields[:-1]
+        else:
+            # If the last field has a register suffix, i.e. "0x38(r7)"  we split that part out to make it a separate field
+            newfields = newfields[:-1] + newfields[-1].split("(")
+            oldfields = oldfields[:-1] + oldfields[-1].split("(")
```
simonlindholm, line 2251: The comment I made before about MIPS is still relevant. MIPS use `%lo(...)` syntax for relocs; if you split on `(` you'll butcher that. This means that e.g.:
- `lui a0, %hi(sym)` vs `lui a0, 0x8000` will split up as `["lui a0", " %hi", "sym)"]` vs `["lui a0", " 0x8000"]`, zip() will ignore the third field of the first, compare " %hi" against " 0x8000" and get a mismatch, and "penalize any extra fields" with add even more to the score. Which is kinda okay as an end result, but the path there is dubious and platform dependent
- `lui a0, %hi(sym)` vs `lui a0, %hi(.rodata + 0x10)` will almost break, because the %hi is split up so that the `imm_matches_everything` logic never finds a `(.`. Except it won't, because `lo_hi_match` takes care of that in advance as an ugly MIPS-specific hack.
- `lui a0, %hi(sym)` vs `lui a1, %hi(.rodata + 0x10)` *will* break, because `lo_hi_match` doesn't treat that case

Basically, this kinda-sorta mostly works, but it's hacky and we're spreading logic between here and `lo_hi_match`, dealing differently with different architectures. There's an opportunity here to get rid of `lo_hi_match` altogether in favor of `imm_matches_everything`. We need to fix this to not catch `(` that are part of %lo/%hi for MIPS though. Ideally we'd have a structured representation of the lines that would make that trivial, but lacking that... how about splitting on a regex that looks for `(` but with a negative lookbehind for `%lo/%hi`?

diff.py
```diff
@@ -2188,7 +2245,16 @@ def diff_sameline(old: str, new: str) -> None:
...
         for nf, of in zip(newfields, oldfields):
+            # If the new field is a match to any symbol case
+            if imm_matches_everything(nf, config.arch):
+                # And the old field had a relocation, then ignore this mismatch
+                if old_line.has_symbol:
+                    continue
```
simonlindholm, line 2257: merge these into a single if joined with an `and`, and move it to inside the `if nf != of` check

diff.py
snuffysasa: okay I did this

diff.py
```diff
@@ -1447,6 +1452,37 @@ def process_reloc(self, row: str, prev: str) -> str:
...
+            # versions depending on if the .o file comes from compiler output or
+            # from disassembly. Normalize, to make sure both forms are treated as
+            # equivalent.
+
+            args = (
+                args.replace("(r2)", "(0)")
+                .replace("(r13)", "(0)")
+                .replace(",r2,", ",0,")
+                .replace(",r13,", ",0,")
+            )
```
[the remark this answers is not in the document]
snuffysasa: ok

diff.py
```diff
@@ -1393,6 +1393,11 @@ class AsmProcessor:
     def __init__(self, config: Config) -> None:
         self.config = config
 
+    def pre_process(
+        self, mnemonic: str, args: str, next_row: Optional[str]
+    ) -> Tuple[str, str]:
+        return (mnemonic, args)
```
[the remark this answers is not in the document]
snuffysasa: black runs everytime I commit... black made it look like this

diff.py
snuffysasa: ohh that is not how I have been using this function, this is probably an error I made a long while back.
This is how I have been using it:


```
        for nf, of in zip(newfields, oldfields):
            # If the new field is a match to any symbol case
            if imm_matches_everything(nf, config.arch):
                # And the old field had a relocation, then ignore this mismatch
                if old_line.has_symbol:
```
 
On a specific code field, and that is why I renamed it, and that is why I made the logic the way I did because it was being used per field.

I will look into resoling this making a separate function, or changing the other places it is used...
the only other place it is used in the coloring code that I haven't done much with there.. maybe it's related to some other color issues

snuffysasa: are there even "@h" in PPC?  I haven't seen any.

vs "@ha" which I do see.

snuffysasa: once this is all finished... I'll circle back to the permuter and make all these fixes/updates there too.

diff.py
snuffysasa: I'll do it, but I'm worried what black will do with it

diff.py
snuffysasa: > Which is kinda okay as an end result, but the path there is dubious and platform dependent

Yes, I analyzed this exact case, and landed that the behavior / end result is the desired behavior.. but I agree the path there is a little dubious.   But won't it result in correct behavior on all platforms?

>lui a0, %hi(sym) vs lui a0, %hi(.rodata + 0x10) will almost break, because the %hi is split up so that the >imm_matches_everything logic never finds a (.

Yea, so I feel like we need correct wildcard symbol catching for each arch... just like for ppc we check for `...data`...  We'd have a similar check here for .rodata + offset, or .data + offset for mips.... I only put it to "(." because I thought I remember a comment you had that said "(." is fine for mips, and I just went with that... maybe I misunderstood.

> will break, because lo_hi_match doesn't treat that case

Same as above.. we just need the correct wildcard symbol catching for MIPS like I worked out for PPC. 

>Basically, this kinda-sorta mostly works, but it's hacky and we're spreading logic between here and lo_hi_match, dealing differently with different architectures. There's an opportunity here to get rid of lo_hi_match altogether in favor of imm_matches_everything.

Yes, makes sense, it makes sense to me get rid of lo_hi_match, and just focus on getting imm_matches_Everything working for all archs..

>We need to fix this to not catch ( that are part of %lo/%hi for MIPS though. Ideally we'd have a structured representation of the lines that would make that trivial, but lacking that... how about splitting on a regex that looks for ( but with a negative lookbehind for %lo/%hi

I am confused why we need this.
In general, I feel like it makes sense to score and check separately for whats in the parenthesis and what is before it. (penalize one or the other, or both that is) And that goes for Mips and for ppc.  We just need to get make a good function for each arch, that detects the wild cards correctly right?

Unless, I am missing something, can you show me a case where splitting at the "(" will not work? assuming we have a good mips function for checking for wildcard symbols.

snuffysasa: Related question

```
    if arch.name == "ppc":
        if "...data" in field:
            return True
```

Do you know if ppc needs to check for more data sections other than just .data... like rodata, sdata, etc... I could never find any examples of where it would.

diff.py
snuffysasa: >function is called with an entire instruction row (well, almost; the branch target part is stripped) by the existing caller.

The only place where this happens is in the row parts coloring section, which I am reading and testing now to understand how it uses it... but I think we will want separate functions for the two places that call this function, since they use it differently.  The other function,  I need to understand and haven't touched is normalize_imms

>normalize_imms(branchless1, arch) == normalize_imms(branchless2, arch)

snuffysasa: yea at first glance, seems the `normalize_imms` is quite broken on ppc, even without any of my changes, or the updated regex.

snuffysasa: So first normalize_imms,
I don't think this can be solved with just a single regex, and I think we will need to break this down and check field by field, and ignore the mnemonics... but you know regex better than me so I'll let you give it a shot. here are some matching examples lhs and rhs.

```
lis     r3,lbl_803CEA7C@ha --- lis     r3,@10@ha
addi    r3,r3,lbl_803CEA7C@l --- addi    r3,r3,@10@l
lis     r3,lbl_803CEA70@ha --- lis     r3,@11@ha
addi    r3,r3,lbl_803CEA70@l --- addi    r3,r3,@11@l
addi    r5,0,lbl_804D3E50@sda21 --- addi    r5,0,@12@sda21
lfs     f0,lbl_804D9A18@sda21(0) --- lfs     f0,@13@sda21(0)

```

Also... now I have it setup so normalize_imms and imm_matches_everything, are the exact same behavior as they were before this PR... and both of these only affect inline coloring only but not scoring....  This is now a separate problem that I will work on and send another PR.  (However it will be slightly improved on ppc, because of the sda21 reloc fixes and li --> addi conversions).  So no need to let this particular issue hold up this current PR, that does the sda21 reloc stuff, and makes some fixes to scoring for all archs.

For this next PR (where I will focus on the inline coloring) I feel like for normalize imms we'll want to change to check this field by field... and same for imm_matches_everything, we will probably want to make that check field by field, or just use the other function I made for scoring.

diff.py
snuffysasa: I am changing mips back to the regex, and now it should have the desired behavior on all the cases you proposed.

snuffysasa: looking at mips code... it seems like any `field` that contains a "." the decimal should be a match to any symbol.
Does that sound right? I don't see an opposing case, but haven't looked at a ton of mips asm.
Because that really simplifies thing if that's the case...

and even if it's not the case, I should be able to mod the code and/or regex to accurately detect.

snuffysasa: and for anything that is not mips and not ppc... it will just return False.. which results in the same behavior as it was before this PR... and will make it easy to add in future for other archs and we want to detect these wildcard(match to any symbol) symbols to improve scoring.

snuffysasa: If you want I can modify this part:
`oldfields = oldfields[:-1] + oldfields[-1].split("(")`
So that it also removes the closing parenthesis at the end. It's not needed for current archs and code, but might be a little more clear/clean

diff.py
snuffysasa: worked out fine

diff.py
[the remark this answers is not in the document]
simonlindholm: Ah, ok. Surprised it doesn't normalize this. Suggest changing to `return mnemonic, args` as a microscopic nit, then (same elsewhere)

diff.py
```diff
@@ -1772,7 +1805,7 @@ class ArchSettings:
     re_reg=re.compile(r"\$?\b([rf][0-9]+)\b"),
     re_sprel=re.compile(r"(?<=,)(-?[0-9]+|-?0x[0-9a-f]+)\(r1\)"),
     re_large_imm=re.compile(r"-?[1-9][0-9]{2,}|-?0x[0-9a-f]{3,}"),
-    re_imm=re.compile(r"(\b|-)([0-9]+|0x[0-9a-fA-F]+)\b(?!\(r1)|[^@]*@(ha|h|lo)"),
+    re_imm=re.compile(r"(\b|-)([0-9]+|0x[0-9a-fA-F]+)\b(?!\(r1)|[^@]*@(ha|h|lo|sda21)"),
```
simonlindholm, line 1808: The `[^@]` here is very dubious btw, given that symbol names can contain `@`. It's probably what causes `normalize_imms` to break. Ideally we wouldn't include symbol names in the asm and then pattern match on that; given compiler-generated names with `@` and C++ mangled names it seems a recipe for failure. Replacing the `[^@]` by some other symbol separator could work though. For later, I guess...

diff.py
```diff
@@ -2033,7 +2077,7 @@ def process(dump: str, config: Config) -> List[Line]:
         num_instr += 1
         source_lines = []
 
-        if config.stop_jrra and mnemonic == "jr" and row_parts[1].strip() == "ra":
+        if config.stop_jrra and mnemonic == "jr" and args.strip() == "ra":
```
simonlindholm, line 2080:
```suggestion
        if config.stop_jrra and mnemonic == "jr" and args == "ra":
```

diff.py
simonlindholm: > are there even "`@h`" in PPC? I haven't seen any.

Not sure if they appear in practice tbh. Extrapolating from MIPS it feels like they'd be used if some code decides to materialize symbol addresses using bitor instead of addition.

diff.py
simonlindholm: %lo(...) is an entire immediate, it doesn't make sense to cut it off halfway, even if we can reason ourselves to concluding that it works okay in practice. Did you look into my suggestion of using a regex with negative lookbehind for %lo/%hi?

diff.py
snuffysasa: >The [^@] here is very dubious btw

I didn't do anything with that or this regex or uses of this regex. I just added the sda21 because you asked me to.

>Replacing the [^@] by some other symbol separator could work though. For later, I guess...

I will start working on this right away, it will be first priority for me (other than fixes to other PRs based on your review)
But it seems like the [^@] is not the main issue, the cases where I see it really breaking do not even include @ symbols.

I'll start digging into and report what I find.

diff.py
simonlindholm: No, never in compiler-generated code. And yeah, I'm hoping we can get rid of lo_hi_match. Not because it gets things wrong, but because it does the same thing as imm_matches_everything but in another manner.

snuffysasa: okay makes sense about getting rid of lo_hi_match.

But you are sure, you don't want to check for all those cases I listed above?

Otherwise I will proceed with updating the code to not split at "(" if there is a proceeding %lo%hi.

And also, we do want to split off and check the final register separately in cases of this right:
%lo(D_808DFEC0)(a1) vs %lo(D_808DFEC0)(a2)

snuffysasa: the negative look-behind for %lo/%hi seems to work... but also I think it would also work, with a negative lookahead (look-forward) for a decimal point, "."

simonlindholm: Yeah, we want to split %lo(D_808DFEC0)(a1) in two. So negative lookahead for `.` wouldn't work

diff.py
```diff
@@ -2251,10 +2251,16 @@ def diff_sameline(old_line: Line, new_line: Line) -> None:
             newfields = newfields[:-1]
             oldfields = oldfields[:-1]
         else:
-            # If the last field has a parenthesis suffix, i.e. "0x38(r7)" or "%lo(.data)" or "%hi(.rodata + 0x10)"
+            # If the last field has a parenthesis suffix, i.e. "0x38(r7)"
             # we split that part out to make it a separate field
-            newfields = newfields[:-1] + newfields[-1].split("(")
-            oldfields = oldfields[:-1] + oldfields[-1].split("(")
+            # however, we don't split if it has a proceeding %hi/%lo  i.e."%lo(.data)" or "%hi(.rodata + 0x10)"
+            if len(oldfields) > 1:
```
simonlindholm, line 2257: no need for this check, from what I can see

diff.py
```diff
@@ -2251,10 +2251,16 @@ def diff_sameline(old_line: Line, new_line: Line) -> None:
...
+                split_last_field = re.split(r"(?<!%hi)(?<!%lo)\(", oldfields[-1])
+                oldfields = oldfields[:-1] + split_last_field
```
simonlindholm, line 2259:
```suggestion
                re_paren = re.compile(r"(?<!%hi)(?<!%lo)\(")
                oldfields = oldfields[:-1] + re_paren.split(oldfields[-1])
```

diff.py
```diff
@@ -2110,7 +2110,7 @@ def field_matches_any_symbol(field: str, arch: ArchSettings) -> bool:
         return re.fullmatch((r"^@\d+$"), field) is not None
 
     if arch.name == "mips":
-        return field[0] == "."
+        return "." in field
```
simonlindholm, line 2113:
```suggestion
        return "(." in field
```
(not that it matters hugely)

diff.py
simonlindholm: `field` -> `row`

diff.py
```diff
@@ -2230,8 +2230,8 @@ def diff_sameline(old_line: Line, new_line: Line) -> None:
         if old == new:
             return
 
-        if lo_hi_match(old, new):
-            return
+        # if lo_hi_match(old, new):
+        #     return
```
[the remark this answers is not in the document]
snuffysasa: sounds good! I just left it for a moment so you can confirm

diff.py
snuffysasa: oh guess you are right

diff.py
snuffysasa: done

diff.py
snuffysasa: leaving how it is, per discord

simonlindholm (approved):

PR #88, opened 2022-06-12
Check each field separately for imms which match everything

snuffysasa:
if they all pass make the whole line white,

If you approve this PR, then you can ignore/close the other, hot fix PR.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2096,10 +2096,6 @@ def normalize_stack(row: str, arch: ArchSettings) -> str:
     return re.sub(arch.re_sprel, "addr(sp)", row)
 
 
-def imm_matches_everything(row_args: str, arch: ArchSettings) -> bool:
-    return "." in row_args
-
-
 def field_matches_any_symbol(field: str, arch: ArchSettings) -> bool:
     if arch.name == "ppc":
         if "..." in field:
@@ -2222,7 +2218,8 @@ def score_diff_lines(
         # Probably regalloc difference, or signed vs unsigned
 
         # Compare each field in order
-        newfields, oldfields = new.split(","), old.split(",")
+        new_parts, old_parts = new.split(None, 1), old.split(None, 1)
+        newfields, oldfields = new_parts[1].split(","), old_parts[1].split(",")
         if ignore_last_field:
             newfields = newfields[:-1]
             oldfields = oldfields[:-1]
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -2441,8 +2438,28 @@ def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
...
+                        if nf != of:
+                            # If the new field is a match to any symbol case
+                            # and the old field had a relocation, then ignore this mismatch
+                            if (
+                                field_matches_any_symbol(nf, config.arch)
+                                and line1.has_symbol
+                            ):
+                                continue
+                            all_fields_match = False
+
```
simonlindholm, line 2461: This is a lot of code, let's put it in a function (i.e. reinstating imm_matches_everything but perhaps with a better name)

diff.py
```diff
@@ -2441,8 +2438,28 @@ def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
...
-                    if len(parts2) > 1 and imm_matches_everything(parts2[1], arch):
+
+                    all_fields_match = True
+                    new_parts = line2.scorable_line.split(None, 1)
+                    old_parts = line1.scorable_line.split(None, 1)
+                    newfields = new_parts[1].split(",")
+                    oldfields = old_parts[1].split(",")
+                    re_paren = re.compile(r"(?<!%hi)(?<!%lo)\(")
+                    oldfields = oldfields[:-1] + re_paren.split(oldfields[-1])
+                    newfields = newfields[:-1] + re_paren.split(newfields[-1])
```
simonlindholm, line 2449: don't duplicate this code

diff.py
```diff
@@ -2441,8 +2438,28 @@ def do_diff(lines1: List[Line], lines2: List[Line], config: Config) -> Diff:
...
+                            # If the new field is a match to any symbol case
+                            # and the old field had a relocation, then ignore this mismatch
+                            if (
+                                field_matches_any_symbol(nf, config.arch)
+                                and line1.has_symbol
+                            ):
+                                continue
+                            all_fields_match = False
+
+                    if all_fields_match:
```
simonlindholm, line 2462: if there's different amounts of fields we also want to treat that as a mismatch

diff.py
```diff
@@ -2222,7 +2218,8 @@ def diff_sameline(old_line: Line, new_line: Line) -> None:
         # Probably regalloc difference, or signed vs unsigned
 
         # Compare each field in order
-        newfields, oldfields = new.split(","), old.split(",")
+        new_parts, old_parts = new.split(None, 1), old.split(None, 1)
+        newfields, oldfields = new_parts[1].split(","), old_parts[1].split(",")
```
simonlindholm, line 2222: can we add a helper function for "split line into mnemonic + arguments"?

diff.py
snuffysasa: sounds good

diff.py
snuffysasa: sounds good

diff.py
snuffysasa: ahh, yep that's right, I meant to include that

diff.py
snuffysasa: sure

PR #91, opened 2022-07-03
Intel i686 architecture

0xfedfedfe:
Worked on two simple relocatable elfs and a Xbox game with symbols.
Loading of Visual C++ style map files isn't complete yet.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1973,7 +1973,6 @@ I686_SETTINGS = ArchSettings(
     ),
     re_large_imm=re.compile(r"-?[1-9][0-9]{2,}|-?0x[0-9a-f]{3,}"),
     re_sprel=re.compile(r"-?(0x[0-9a-f]+|[0-9]+)(?=\((%ebp|%esi)\))"),
-    # (?<=\$)
     re_imm=re.compile(r"-?(0x[0-9a-f]+|[0-9]+)"),
     re_reloc=re.compile(r"R_386_"),
     arch_flags=["-m", "i386"],
--- a/diff.py
+++ b/diff.py
@@ -1070,6 +1070,7 @@ def preprocess_objdump_out(
 
     return out
 
+
 def search_build_objects(objname: str, project: ProjectSettings) -> Optional[str]:
     objfiles = [
         os.path.join(dirpath, f)
@@ -1087,6 +1088,7 @@ def search_build_objects(objname: str, project: ProjectSettings) -> Optional[str
         return objfiles[0]
 
     return None
+
 
 def search_map_file(
     fn_name: str, project: ProjectSettings, config: Config
@@ -1174,7 +1176,7 @@ def search_map_file(
         if not diff_segment_find:
             fail(f"Couldn't find segment for section in map file.")
         diff_segment = diff_segment_find.group(1)
-        
+
         find = re.findall(
             re.compile(
                 r" (?:"
@@ -1189,7 +1191,11 @@ def search_map_file(
             fail(f"Found multiple occurrences of function {fn_name} in map file.")
         if len(find) == 1:
             names_find = re.search(r"(\S+) ... (\S+)", find[0])
-            fileofs = int(names_find.group(1), 16) - load_address + project.ms_map_address_offset
+            fileofs = (
+                int(names_find.group(1), 16)
+                - load_address
+                + project.ms_map_address_offset
+            )
             objname = names_find.group(2)
             objfile = search_build_objects(objname, project)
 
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -1606,6 +1638,49 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
...
+        elif "R_386_GOTOFF" in row:
+            repl = f"%got({repl})"
+        elif "R_386_GOTPC" in row:
+            repl = f"%got({repl})"
+        elif "R_386_32PLT" in row:
+            repl = f"%plt({repl})"
+        else:
+            assert False, f"unknown relocation type '{row}' for line '{prev}'"
+
+        return f"{mnemonic}\t{repl}{args_string}", repl
```
simonlindholm, line 1681: the relocated part of the instruction doesn't have to be the first one: https://godbolt.org/z/z6fhMvzse (x86_64, but I think x86 is the same)

diff.py
```diff
@@ -1730,6 +1805,76 @@ class ArchSettings:
     "bgt-",
 }
 
+I686_BRANCH_INSTRUCTIONS = {
+    "call",
```
simonlindholm, line 1809: mips doesn't include its function call instruction here fwiw. I think currently we might crash if a "branch instruction" has a relocation, because we try to parsed the relocated symbol as a base 16 address? which would be possible to fix ofc, just something you might run into

diff.py
```diff
@@ -1609,6 +1674,62 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
...
+        elif "R_386_GOTOFF" in row:
+            repl = f"%got({repl})"
+        elif "R_386_GOTPC" in row:
+            repl = f"%got({repl})"
+        elif "R_386_32PLT" in row:
+            repl = f"%plt({repl})"
+        else:
+            assert False, f"unknown relocation type '{row}' for line '{prev}'"
+
+        if reverse:
```
simonlindholm, line 1727: rather than a `reverse` variable I think it'd be cleaner to have `before` and `after`, similar to reloc code for most other arches

diff.py
```diff
@@ -1609,6 +1674,62 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
...
+
+        # Register relative addressed target e.g.
+        # R_386_GOTOFF        x
+        # movl      $0x1,0x0(%eax)
+        if comma != -1:
+            if open_bracket != -1 and open_bracket > comma:
+                reverse = True
+                args_string = args[: comma + 1]
+            else:
+                args_string = args[comma:]
```
simonlindholm, line 1696: This logic is incomplete:
```c
int x[2], y[2];

int foo() {
	y[1] = 1;
	return x[1];
}
```
compiled with `gcc test.c -c -m32 -O2 -fno-pic && objdump --no-show-raw-in -dr test.o` yields
```
00000000 <foo>:
   0:	movl   $0x1,0x4
			2: R_386_32	y
   a:	mov    0x4,%eax
			b: R_386_32	x
   f:	ret    
```

`movl   $0x1,0x4` has the relocated symbol as second argument, while this code assumes the first.

This also shows off an interesting property of GCC-compiled x86 ELFs, in that GCC chooses to use REL relocations, with addend embedded into the asm. (On x86_64 it uses RELA relocs where they are separate, resulting in the equivalent of `movl   $0x1,0x0` / `R_386_32   y+4` instead). I don't know how your toolchain looks for MSVC compilation but I imagine RELA is more likely so no need to worry about this I guess.

Also, I'm not sure stripping parentheses entirely is the right move. At least for MIPS it would result in invalid asm, and it loses information. Example:
```c
extern int x[20], y[20];

int foo(int a) {
	y[a + 1] = 1;
	return x[a + 2];
}
```

```
00000000 <foo>:
   0:	mov    0x4(%esp),%eax
   4:	movl   $0x1,0x4(,%eax,4)
			7: R_386_32	y
   f:	mov    0x8(,%eax,4),%eax
			12: R_386_32	x
  16:	ret    
```

diff.py
```diff
@@ -445,7 +447,9 @@ def create_project_settings(settings: Dict[str, Any]) -> ProjectSettings:
         objdump_executable=get_objdump_executable(settings.get("objdump_executable")),
         objdump_flags=settings.get("objdump_flags", []),
         map_format=settings.get("map_format", "gnu"),
-        mw_build_dir=settings.get("mw_build_dir", "build/"),
+        ms_map_address_offset=settings.get("ms_map_address_offset", 0),
+        ms_ignore_missing_objfile=settings.get("ms_ignore_missing_objfile", True),
+        build_dir=settings.get("build_dir", "build/"),
```
simonlindholm, line 452:
```suggestion
        build_dir=settings.get("build_dir", settings.get("mw_build_dir", "build/")),
```
to avoid hassle for mwcc decomp projects

simonlindholm (commented):
Sorry for the review delay. I made some remarks (about relocation handling mostly), but I'm also happy to merge stuff in an incomplete state if you don't want to deal with it, just lmk if so.

Issue #69, opened 2021-12-23
Properly handle R_PPC_REL14


mkst:
via https://github.com/decompme/decomp.me/issues/220

simonlindholm:
having the full objdump output, or even better a crashing .o file, would be useful. Also having target asm for that function, if it hints at how to write a macro for this relocation

ethteck:
[simpleStart__12RumbleSampleFf.zip](https://github.com/simonlindholm/asm-differ/files/7807349/simpleStart__12RumbleSampleFf.zip)

simonlindholm:
This happens because the .s file contains `bne lbl_8017D700` where lbl_8017D700 is not a local label. Defining it as one fixes the error and has a bunch of other advantages like making the asm shiftable and getting rid of diff noise (that doesn't appear as diffs in the final rom but is a diff at elf level).

We should still try not to crash here though...

mkst:
Can/should/could this be closed?

ethteck:
It appears to be fixed, at least according to the linked decomp.me scratch for the corresponding issue there

simonlindholm:
Not totally sure we handle R_PPC_REL14 *correctly* (which was why I didn't close this previously), but yeah, it's probably best not to leave this issue around forever, let's reopen if someone actually complains.

PR #92, opened 2022-07-03
PPC_REL14 fix

snuffysasa:
fix that's making scratches crash that have this unknown relocation: R_PPC_REL14


simonlindholm:
Sure, seems fine. #69 has a bit of discussion on this.

PR #93, opened 2022-07-07
ppc updates

SeekyCt:
- Made `--stop-jr-ra` work in ppc for the `blr` instruction
- Allowed multi-digit alignments in the mw symbol map regex
- Added arch_flags for ppc
    - `-m powerpc` fixes objdump saying "can't disassemble for architecture UNKNOWN!" when using on a biniary
    - `-M gekko` fixes disassembly of paired single instructions

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1112,7 +1112,7 @@ def search_map_file(
         find = re.findall(
             re.compile(
                 #            ram   elf rom
-                r"  \S+ \S+ (\S+) (\S+)  . "
+                r"  \S+ \S+ (\S+) (\S+) +\S+ "
                 + re.escape(fn_name)
                 + r"(?: \(entry of "
                 + re.escape(config.diff_section)
--- a/diff.py
+++ b/diff.py
@@ -204,8 +204,8 @@ if __name__ == "__main__":
         "--stop-jr-ra",
         dest="stop_jrra",
         action="store_true",
-        help="""Stop disassembling at the first 'jr ra'. Some functions have
-        multiple return points, so use with care!""",
+        help="""Stop disassembling at the first 'jr ra' (mips) or 'blr' (ppc).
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -204,8 +204,8 @@ def complete_symbol(
         "--stop-jr-ra",
         dest="stop_jrra",
         action="store_true",
-        help="""Stop disassembling at the first 'jr ra'. Some functions have
-        multiple return points, so use with care!""",
+        help="""Stop disassembling at the first 'jr ra' (mips) or 'blr' (ppc).
```
simonlindholm, line 207:
```suggestion
        help="""Stop disassembling at the first return instruction.
```

diff.py
```diff
@@ -204,8 +204,8 @@ def complete_symbol(
         "--stop-jr-ra",
```
simonlindholm, line 204: can we rename this --stop-at-ret?

diff.py
```diff
@@ -1813,6 +1813,7 @@ class ArchSettings:
         r"(\b|-)([0-9]+|0x[0-9a-fA-F]+)\b(?!\(r1)|[^ \t,]+@(l|ha|h|sda21)"
     ),
     re_reloc=re.compile(r"R_PPC_"),
+    arch_flags=["-m", "powerpc", "-M", "gekko"],
```
simonlindholm, line 1816:
```suggestion
    arch_flags=["-m", "powerpc", "-M", "broadway"],
```
to match permuter; I think behavior should be identical

diff.py
```diff
@@ -1112,7 +1112,7 @@ def search_map_file(
         find = re.findall(
             re.compile(
                 #            ram   elf rom
```
simonlindholm, line 1114:
```suggestion
                #            ram   elf rom  alignment
```

simonlindholm (approved):

Issue #94, opened 2022-07-08
Line numbers break with mwcceppc dwarf-2


SeekyCt:
When codewarrior is passed the `-sym dwarf-2` flag, a full path is output for the source file name, which includes a colon as part of the drive name and therefore breaks this regex https://github.com/simonlindholm/asm-differ/blob/b0ef0c5bf47573c5b1cb76b1627c834d806815e1/diff.py#L1958

Would allowing colons here break other things?

simonlindholm:
I think it should be fine to allow.

SeekyCt:
Cool, thanks

PR #95, opened 2022-07-08
Allow colons in source line number regex

SeekyCt:
Fix for #94


diff.py
```diff
@@ -1955,7 +1955,7 @@ def process(dump: str, config: Config) -> List[Line]:
             # This regex is conservative, and assumes the file path does not contain "weird"
             # characters like colons, tabs, or angle brackets.
```
simonlindholm, line 1956:
```suggestion
            # characters like tabs or angle brackets.
```

simonlindholm (approved):

diff.py
SeekyCt: Oh yeah, whoops

PR #97, opened 2022-11-07
add support of french

coco875:
add support of french

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1074,6 +1074,10 @@ def preprocess_objdump_out(
 
     return out
 
+load_address_lang = [
+    "load address",
+    "address de chargement",
+]
 
 def search_build_objects(objname: str, project: ProjectSettings) -> Optional[str]:
     objfiles = [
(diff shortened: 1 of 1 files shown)


AngheloAlf:
We had a similar problem on MM and it was fixed by adding this to the makefile:
```make
# Ensure the map file being created using English localization
export LANG := C
```
Basically it ensures the files are created in English localization instead of the system's language

coco875:
I don't if it's better to change make or update the tool but add in every language to support this

simonlindholm:
IMO changing asm-differ to support every language feels error prone (what if new translations get added, or existing ones change?), and there are other tools that parse map files too, and it would be annoying to add support to all of them. Can we best-effort let asm-differ print an error message about building with LANG = C if it encounters a non-English string like this?

coco875:
I add a assert to make a error when he don't find load address in file and add the advice to add 'export LANG := C' in the makefile

coco875:
and remove what I made

simonlindholm:
Nice approach with the substring search, in my mind it was more complex. I had some comments but it felt easier to just address them myself, so I committed 1e81f181396b9561d0ec5b5329b609068a74b348. Thanks!

PR #99, opened 2022-12-09
One-way and zero-way diffing, and expected_directory config option

### LINKED ISSUE #49: Support viewing one "side" only and not doing a diff
Allow accepting only one input and just displaying it back with no diff calculation

ethteck:
I've added two new cli args in this PR: `-0` and `-1` for just viewing (not diffing) asm by itself. The 0 mode views base and 1 views current.

This is currently done by just telling asm-differ to diff one side against itself and then only display one of the two identical sides. I thought this would be the best way to go about doing this, as it allows us to run through all existing diff code (which probably tidies things up as well) and not have to touch too many different parts of the codebase in doing so, but I'm happy to try going a different direction if people have suggestions.

I also added a config option to specifiy the expected directory, as Paper Mario has a non-standard directory structure which previously required dumb manual modifications to diff.py - no longer! (fixes #49)

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -3100,13 +3100,11 @@ def align_diffs(old_diff: Diff, new_diff: Diff, config: Config) -> TableData:
         diff_lines = compress_matching(diff_lines, config.compress.context)
 
     def diff_line_to_table_line(line: Tuple[OutputLine, ...]) -> TableLine:
-        cells: List[Tuple[Text, Optional["Line"]]] = []
-        for i in range(0, len(line)):
-            if i == 0:
-                base: Text = line[i].base or Text()
-                cells.append((base, line[i].line1))
-            else:
-                cells.append((line[i].fmt2, line[i].line2))
+        cells: List[Tuple[Text, Optional["Line"]]] = [
+            (line[0].base or Text(), line[0].line1)
+        ]
+        for i in range(1, len(line)):
+            cells.append((line[i].fmt2, line[i].line2))
 
         return TableLine(
             key=line[0].key2,
--- a/diff.py
+++ b/diff.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # PYTHON_ARGCOMPLETE_OK
 import argparse
+import enum
 import sys
 from typing import (
     Any,
@@ -26,6 +27,14 @@ def fail(msg: str) -> NoReturn:
 
 def static_assert_unreachable(x: NoReturn) -> NoReturn:
     raise Exception("Unreachable! " + repr(x))
+
+
+class DiffMode(enum.Enum):
+    SINGLE = "single"
+    SINGLE_BASE = "single_base"
+    NORMAL = "normal"
+    THREEWAY_PREV = "3prev"
+    THREEWAY_BASE = "3base"
 
 
 # ==== COMMAND-LINE ====
@@ -257,35 +266,35 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "-0",
-        "--diff_type=single_base",
-        dest="diff_type",
+        "--diff_mode=single_base",
+        dest="diff_mode",
         action="store_const",
-        const="single_base",
+        const=DiffMode.SINGLE_BASE,
         help="""View the base asm only (not a diff).""",
     )
     parser.add_argument(
         "-1",
-        "--diff_type=single",
-        dest="diff_type",
+        "--diff_mode=single",
+        dest="diff_mode",
         action="store_const",
-        const="single",
+        const=DiffMode.SINGLE,
         help="""View the current asm only (not a diff).""",
     )
     parser.add_argument(
         "-3",
-        "--diff_type=3prev",
-        dest="diff_type",
+        "--threeway=prev",
+        dest="diff_mode",
         action="store_const",
-        const="3prev",
+        const=DiffMode.THREEWAY_PREV,
         help="""Show a three-way diff between target asm, current asm, and asm
         prior to -w rebuild. Requires -w.""",
     )
     parser.add_argument(
         "-b",
-        "--diff_type=3base",
-        dest="diff_type",
+        "--threeway=base",
+        dest="diff_mode",
         action="store_const",
-        const="3base",
+        const=DiffMode.THREEWAY_BASE,
         help="""Show a three-way diff between target asm, current asm, and asm
         when diff.py was started. Requires -w.""",
     )
@@ -363,7 +372,6 @@ import ast
 from collections import Counter, defaultdict
 from dataclasses import asdict, dataclass, field, replace
 import difflib
-import enum
 import html
 import itertools
 import json
@@ -418,14 +426,6 @@ class Compress:
     same_instr: bool
 
 
-class DiffType(enum.Enum):
-    SINGLE = "single"
-    SINGLE_BASE = "single_base"
-    NORMAL = "normal"
-    THREEWAY_PREV = "3prev"
-    THREEWAY_BASE = "3base"
-
-
 @dataclass
 class Config:
     arch: "ArchSettings"
@@ -442,7 +442,7 @@ class Config:
 
     # Display options
     formatter: "Formatter"
-    diff_type: DiffType
+    diff_mode: DiffMode
     base_shift: int
     skip_lines: int
     compress: Optional[Compress]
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -418,6 +433,8 @@ class Config:
     # Display options
     formatter: "Formatter"
     threeway: Optional[str]
+    zeroway: bool
+    oneway: bool
```
simonlindholm, line 437: combine with `threeway` here too, possibly use an `Enum`

diff.py
```diff
@@ -255,6 +255,20 @@ def complete_symbol(
         help="""Automatically update when source/object files change.
         Recommended in combination with -m.""",
     )
+    parser.add_argument(
+        "-0",
+        "--zeroway",
+        dest="zeroway",
```
simonlindholm, line 261: this can use the same dest as -b/-3 with another `const`, since it's in the same category as those (and incompatible with them)

diff.py
```diff
@@ -255,6 +255,20 @@ def complete_symbol(
         help="""Automatically update when source/object files change.
         Recommended in combination with -m.""",
     )
+    parser.add_argument(
+        "-0",
+        "--zeroway",
```
simonlindholm, line 260: The term zeroway is a bit silly, maybe --oneway-base? The term oneway is also misleading and would imo be good to replace with something else, though I don't have a great suggestion off hand.

diff.py
```diff
@@ -1411,7 +1435,7 @@ def dump_objfile(
     if not os.path.isfile(objfile):
         fail(f"Not able to find .o file for function: {objfile} is not a file.")
 
-    refobjfile = "expected/" + objfile
+    refobjfile = project.expected_directory + objfile
```
simonlindholm, line 1438:
```suggestion
    refobjfile = os.path.join(project.expected_directory, objfile)
```
(which will make it work even if a trailing / is missing)

diff.py
```diff
@@ -3049,6 +3073,14 @@ def align_diffs(
         diff_lines = [
             (base, new, old if old != new else empty) for base, new, old in diff_lines
         ]
+    elif config.zeroway or config.oneway:
+        meta = TableMetadata(
+            headers=(Text("CURRENT"),),
```
simonlindholm, line 3078: maybe it should show BASE for zeroway

diff.py
```diff
@@ -3073,9 +3079,11 @@ def align_diffs(
         diff_lines = [
             (base, new, old if old != new else empty) for base, new, old in diff_lines
         ]
-    elif config.zeroway or config.oneway:
+    elif config.diff_type in [DiffType.SINGLE or DiffType.SINGLE_BASE]:
```
simonlindholm, line 3082:
```suggestion
    elif config.diff_type in (DiffType.SINGLE, DiffType.SINGLE_BASE):
```
(() or {} being slightly more idiomatic than [] I think? but I don't remember where I got that from)

diff.py
```diff
@@ -3073,9 +3079,11 @@ def align_diffs(
...
         meta = TableMetadata(
-            headers=(Text("CURRENT"),),
+            headers=(
+                Text("BASE" if config.diff_type == DiffType.SINGLE_BASE else "CURRENT"),
+            ),
```
simonlindholm, line 3086:
```suggestion
            headers=(Text(header),),
```
+ local variable, to avoid the ugly wrapping

diff.py
```diff
@@ -3198,15 +3206,15 @@ def run_diff(self) -> Tuple[str, object]:
...
             diff_output = do_diff(self.base_lines, self.base_lines, self.config)
-        elif self.config.oneway:
+        elif self.config.diff_type == DiffType.SINGLE:
             diff_output = do_diff(my_lines, my_lines, self.config)
         else:
             diff_output = do_diff(self.base_lines, my_lines, self.config)
 
         last_diff_output = self.last_diff_output or diff_output
-        if self.config.threeway != "base" or not self.last_diff_output:
+        if self.config.diff_type != "base" or not self.last_diff_output:
```
simonlindholm, line 3217: use the enum here

diff.py
```diff
@@ -524,9 +532,7 @@ def create_config(args: argparse.Namespace, project: ProjectSettings) -> Config:
         max_function_size_bytes=args.max_lines * 4,
         # Display options
         formatter=formatter,
-        threeway=args.threeway,
-        zeroway=args.zeroway,
-        oneway=args.oneway,
+        diff_type=DiffType(args.diff_type or DiffType.NORMAL),
```
simonlindholm, line 535: can the argparse `const` values be enum members already, so you don't have to do this fallible str -> enum conversion?

diff.py
```diff
@@ -416,6 +418,14 @@ class Compress:
     same_instr: bool
 
 
+class DiffType(enum.Enum):
```
simonlindholm, line 421: Mode could be clearer than Type. It's still imperfect, because there's multiple modes being configured -- also e.g. -o vs --elf vs neither -- but whatever, it's not user-exposed and the usage is clear from context.

diff.py
```diff
@@ -257,33 +257,35 @@ def complete_symbol(
...
+        dest="diff_type",
+        action="store_const",
+        const="single",
         help="""View the current asm only (not a diff).""",
     )
     parser.add_argument(
         "-3",
-        "--threeway=prev",
-        dest="threeway",
+        "--diff_type=3prev",
```
simonlindholm, line 276: not a fan of the snake casing, internal-looking "3prev" token, and breaking backward compat by changing this

diff.py
ethteck: not sure why I put "or" there..

diff.py
```diff
@@ -3079,29 +3076,66 @@ def align_diffs(
...
+                    base: Text = line[i].base  # type: ignore
+                    cells.append((base, line[i].line1))
+                else:
+                    cells.append((line[i].fmt2, line[i].line2))
+
+            assert line[0].key2 is not None
+            table_lines.append(
+                TableLine(
+                    key=line[0].key2,
+                    is_data_ref=line[0].is_data_ref,
```
simonlindholm, line 3126: (we were using [1] before, and now [0]. I think they're probably equally correct but I'll think a little more about this)

diff.py
```diff
@@ -455,6 +480,7 @@ def create_project_settings(settings: Dict[str, Any]) -> ProjectSettings:
         ),
         objdump_executable=get_objdump_executable(settings.get("objdump_executable")),
         objdump_flags=settings.get("objdump_flags", []),
+        expected_directory=settings.get("expected_directory", "expected/"),
```
simonlindholm, line 483: can you add this commented-out to diff_settings.py? and maybe rename it "expected_dir" for consistency with "build_dir"?

diff.py
```diff
@@ -3049,20 +3071,65 @@ def align_diffs(
...
+                )
+            )
+        return table_lines
+
+    return TableData(
+        headers=headers,
+        current_score=current_score,
+        max_score=max_score,
+        previous_score=previous_score,
+        lines=diff_lines_to_table_lines(diff_lines),
```
simonlindholm, line 3131:
```suggestion
        lines=[diff_line_to_table_line(line) for line in diff_lines],
```
to avoid the need for a loop in the function

diff.py
```diff
@@ -3049,20 +3071,65 @@ def align_diffs(
...
+        diff_lines: List[Tuple[OutputLine, ...]]
+    ) -> List[TableLine]:
+        table_lines: List[TableLine] = []
+        for line in diff_lines:
+
+            cells: List[Tuple[Text, Optional["Line"]]] = []
+            for i in range(0, len(line)):
+                if i == 0:
+                    assert line[i].base is not None
+                    base: Text = line[i].base  # type: ignore
```
simonlindholm, line 3112: fairly sure this isn't right and you need an `or Text()` (grep for `base=None`)

diff.py
```diff
@@ -3049,20 +3071,65 @@ def align_diffs(
...
+
+    def diff_lines_to_table_lines(
+        diff_lines: List[Tuple[OutputLine, ...]]
+    ) -> List[TableLine]:
+        table_lines: List[TableLine] = []
+        for line in diff_lines:
+
+            cells: List[Tuple[Text, Optional["Line"]]] = []
+            for i in range(0, len(line)):
+                if i == 0:
```
simonlindholm, line 3110: https://thedailywtf.com/articles/Switched_on_Loops

simonlindholm (approved):

PR #102, opened 2023-03-09
Change ast eval to plain eval

ethteck:
Nim found an issue in that you can't pass an expression to the `base_shift` argument. It turns out `ast.literal_eval` doesn't support a string containing operations. Eval does, so I switched to it. Maybe there's a smarter thing that can be done, but I think this is pretty harmless.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -368,7 +368,6 @@ if __name__ == "__main__":
 # (We do imports late to optimize auto-complete performance.)
 
 import abc
-import ast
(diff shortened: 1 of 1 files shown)


simonlindholm:
Makes me vaguely uncomfortable and it would be nice if there was a built-in way to evaluate just arithmetic, but yeah, I suppose it's fine and there's likely other channels for RCE if you can pass arbitrary cmdline args.

PR #106, opened 2023-04-06
Basic SH2 support

sozud:
This PR adds basic support for the Hitachi SH2 architecture. I'm not sure about relocations, re_sprel, or re_large_imm. The binary I'm testing on seems to work ok without relocations. The SH2 doesn't have stack pointer relative instructions and the largest immediate size is a 12-bit displacement for BRA and BSR. Otherwise it's 8 bit. Instruction list is on PDF page 44 here: https://antime.kapsi.fi/sega/files/h12p0.pdf

Highlighting and branches seem to be OK by my testing:

image

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -557,6 +557,7 @@ def get_objdump_executable(objdump_executable: Optional[str]) -> str:
         "mips-linux-gnu-objdump",
         "mips64-elf-objdump",
         "mips-elf-objdump",
+        "sh-elf-objdump",
     ]
     for objdump_cand in objdump_candidates:
         try:
@@ -1811,6 +1812,17 @@ class AsmProcessorI686(AsmProcessor):
         return mnemonic == "ret"
 
 
+class AsmProcessorSH2(AsmProcessor):
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
+
+    def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
+        return prev, None
+
+    def is_end_of_function(self, mnemonic: str, args: str) -> bool:
+        if mnemonic == "rts":
+            return True
+
 @dataclass
 class ArchSettings:
     name: str
@@ -2005,6 +2017,15 @@ I686_BRANCH_INSTRUCTIONS = {
     "jz",
 }
 
+SH2_BRANCH_INSTRUCTIONS = {
+    "bf",
+    "bf.s",
+    "bt",
+    "bt.s",
+    "bra",
+    "bsr"
+}
+
 MIPS_SETTINGS = ArchSettings(
     name="mips",
     re_int=re.compile(r"[0-9]+"),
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -2135,6 +2179,7 @@ class ArchSettings:
     AARCH64_SETTINGS,
     PPC_SETTINGS,
     I686_SETTINGS,
+    SH2_SETTINGS
```
simonlindholm, line 2182:
```suggestion
    SH2_SETTINGS,
```

diff.py
```diff
@@ -2005,6 +2017,15 @@ class ArchSettings:
     "jz",
 }
 
+SH2_BRANCH_INSTRUCTIONS = {
+    "bf",
+    "bf.s",
+    "bt",
+    "bt.s",
+    "bra",
+    "bsr"
```
simonlindholm, line 2026:
```suggestion
    "bsr",
```

simonlindholm (approved):
Looks ok! Relocs ought to appear if you reference global variables and diff in -o mode, and are super useful: it makes it much easier to understand the asm. Definitely not something that needs to block this PR though. Can you add a mention of the arch in the README and argparse help text?

sozud:
Ok, all the relocs I have in my current file seem to be equal to 0 so I guess it's a no-op? For example this excerpt:

```
 Offset     Info    Type            Sym.Value  Sym. Name + Addend
0000027c  00001501 R_SH_DIR32        00000000   ___sdivsi3 + 0
00000660  00001c01 R_SH_DIR32        00000000   _msx + 0
00000674  00001a01 R_SH_DIR32        00000000   ___ashrsi3 + 0
000006bc  00000301 R_SH_DIR32        00000000   .data + 0
000006c0  00001b01 R_SH_DIR32        00000524   _Hw32xScreenPutChar + 0
00000898  00002101 R_SH_DIR32        00000000   _vsnprintf + 0
```

I think I'll come back to that part once I find an example that's not working. I applied the requested changes to argparse and the readme.

simonlindholm:
Those relocs aren't to 0, they are equal to someSymbol + 0. someSymbol will be resolved to some concrete memory address during the linking step.

diff.py supports a few different modes, chiefly, -o and non- -o (binary diff). In the binary diff mode you diff linked files, and get asm that looks like `lui $a0, 0x1234`, `lw $a0, 0x5678($a0)` (or however that looks in SH4) from C code like `int a = someGlobal;`. In -o mode, you instead get `lui $a0, %hi(someGlobal)`, `lw $a0, %lo(someGlobal)($a0)`. I.e., relocs let you see symbol names. If you have C code like `int a = someGlobal.structMemberAtOffset4` you get `lui $a0, %hi(someGlobal + 4)`, `lw $a0, %lo(someGlobal + 4)($a0)`, which is a case where you wouldn't see + 0 above.

Issue #113, opened 2023-06-19
SH2 Improvements


sozud:
I'm running into some issues with data making decompiling SH2 annoying. With SH2, there's only 8-bit immediates, so most data is loaded with pc-relative instructions, and the data is interspersed with the function. This scratch is an example of the sort of issues: https://decomp.me/scratch/SUNET

Lines 18-20 in the asm is actually a jump table.
The 0xB0 in the switch is at line 5A in the asm.
Lines 5C-5E in the asm is the pointer to D_800A7734.

Longer functions have this issue worse since the pc-relative offset is limited so there will be a block instructions and a jump, a block of data, the next block of instructions and a jump, a block of data, etc.

I've been thinking about different ways to solve this issue and was wondering if anyone has suggestions. Do any other architectures have this sort of problem?

I've though about:

Replace objdump with a better disassembler for the SH2 case? My disassembler has gnu as-compatible output but it's written in rust so that would be a significant dependency.
Add more parsing to asm-differ to try the make the objdump output better? I'm not sure exactly how much additional parsing is needed but it seems like it would be basically re-implementing a disassembler for certain patterns.
(Not really asm-differ related) Allow linker arguments in decomp.me so that the pointers can be set to the right locations? This would help with cases like struct->offset where the offset and the base pointer added together by gcc.

simonlindholm:
> I've been thinking about different ways to solve this issue and was wondering if anyone has suggestions. Do any other architectures have this sort of problem?

I think arm does, but its offsets reach further so it's less of a problem. Haven't used it myself though.

Switching to a different disassembler seems fine to me as long as it's an optional dependency, and as long as the different syntax doesn't end up bloating the code too much by requiring SH2-specific handling everywhere. (So some amount of objdump mimicry might be required? Or we could change the code to separate instruction parsing from diffing, having all other arches use regexes just for instruction parsing instead of the current mess, but that would be a much larger change.)

I wonder if objdump could be made into emitting .word for all literal pools by marking those parts of the .o as not ST_FUNC. Maybe some tool could fix up .o files just before diffing or just after compilation?

>  (Not really asm-differ related) Allow linker arguments in decomp.me so that the pointers can be set to the right locations? This would help with cases like struct->offset where the offset and the base pointer added together by gcc.

[...]

PR #114, opened 2023-06-26
Add SH4 support

EstexNT:
This PR adds basic Hitachi SH4 support. Based off @sozud's work on SH2 which means it also relates to #113 

image

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1826,17 +1826,6 @@ class AsmProcessorSH2(AsmProcessor):
         return mnemonic == "rts"
 
 
-class AsmProcessorSH4(AsmProcessor):
-    def __init__(self, config: Config) -> None:
-        super().__init__(config)
-    
-    def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
-        return prev, None
-    
-    def is_end_of_function(self, mnemonic: str, args: str) -> bool:
-        return mnemonic == "rts"
-
-
 @dataclass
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -2177,6 +2189,20 @@ class ArchSettings:
...
+SH4_SETTINGS = replace(
+    SH2_SETTINGS, 
+    name="sh4",
+    #   - fr0-fr15, dr0-dr14, xd0-xd14, fv0-fv12 FP registers
+    #     dr/xd registers can only be even-numbered, and fv registers can only be a multiple of 4
+    re_reg=re.compile(
+        r"r1[0-5]|r[0-9]|fr1[0-5]|fr[0-9]|dr[02468]|dr1[024]|xd[02468]|xd1[024]|fv[048]|fv12"
+    ),
+    arch_flags=["-m", "sh4"], 
+    proc=AsmProcessorSH4,
```
simonlindholm, line 2201: maybe we should use AsmProcessorSH2 instead of defining a new one? I don't know enough about the arch to say whether there are any cases where we have to treat them differently, so I'll let you decide

simonlindholm (approved):

diff.py
EstexNT: AFAIK there aren't (and won't be) any differences between the two AsmProcessors -- I'll remove AsmProcessorSH4 then

sozud (approved):

PR #117, opened 2023-08-11
Add m68k support

asdf2jkl:
Add support for the 68000 processor. Support for later processors in the series is untested and unintended. Overall the adaptation is straightforward, with only two very minor complications:

1) objdump uses a downright bizzare syntax for memory accesses which is
   unlike that used in any assemblers, so I re-format those accesses
   with their equivilant in AT&T syntax using an ugly regex.

2) Nearly every instruction on the m68k can take an immediate address
   operand and most (but not all) instructions have several suffixes
   denoting their width, so generating the set of instructions that can
   take address immediates gets complicated.

This patch has been tested with both raw binary files and ELF object files, and with both objdump 2.16.1 and 2.40.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -54,7 +54,7 @@ if __name__ == "__main__":
         argcomplete = None
 
     parser = argparse.ArgumentParser(
-        description="Diff MIPS, PPC, AArch64, ARM32, SH2, or SH4 assembly."
+        description="Diff MIPS, PPC, AArch64, ARM32, SH2, SH4, or m68k assembly."
     )
 
     start_argument = parser.add_argument(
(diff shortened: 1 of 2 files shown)


asdf2jkl:
I should note that the generation of `M68K_INSTRUCTIONS_WITH_ADDRESS_IMMEDIATES` creates a number of mnemonics with invalid suffixes. Not all instructions can have all suffixes, but since objdump won't emit these bogus instructions, there's no problem with having them in the set.

diff.py
```diff
@@ -1826,6 +1827,57 @@ def is_end_of_function(self, mnemonic: str, args: str) -> bool:
...
+            r"\5\2(%\1\3)\4",
+            args,
+        )
+
+    def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
+        repl = row.split()[-1]
+        mnemonic, args = prev.split(maxsplit=1)
+
+        print(prev + "\n" + row)
+
```
simonlindholm, line 1846:
```suggestion
```

simonlindholm (approved):

PR #118, opened 2023-08-11
Assorted m68k fixes/enhancements

asdf2jkl:
It turns out that the m68k instruction set is too orthogonal for the `instructions_with_address_immediates` to be useful. As most instructions _can_ take an address immediate, I listed them all. But it is more common for the second operand to be a register, which is not detected when the instruction is in the `instructions_with_address_immediates` set. Remove all possible false positive mnemonics.

Fixes:
 - This allows proper detection of register changes when they are the second operand.
 - Diffs in sp-offsets are now detected correctly.
 
Adds:
 - Detects ccr as a register
 - Detects all 68881 floating-point and control registers

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2111,78 +2111,6 @@ M68K_BRANCH_INSTRUCTIONS = {
     }
 )
 
-M68K_INSTRUCTIONS_WITH_ADDRESS_IMMEDIATES_AND_SUFFIXES = {
-    "add",
-    "adda",
-    "addi",
-    "addq",
-    "and",
-    "andi",
-    "asl",
-    "asr",
-    "bchg",
-    "bclr",
-    "bset",
-    "btst",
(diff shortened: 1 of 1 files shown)


ethteck:
Could I ask what project(s) motivate this work? I'm interested in seeing what you're doing with this processor and asm diffing 👁️

ethteck:
that's so cool! Let us know when you get it all set up and public, if you don't mind. We could maybe look into adding platform support on decomp.me as well, if you're interested

asdf2jkl:
I can totally let you guys know when I open it up. I'm planning on getting the memory management and some link port handling done before I do so.

I don't know the decomp.me ecosystem, but I'm afraid that my project won't be suitable because of the license that the compiler falls under. I may take a look at adding general m68k support if I have the time.

PR #119, opened 2023-08-13
Support -f flag when in binary diff mode

asdf2jkl:
Currently you can only specify a file to be diffed on the command line when you are diffing object files.

My bespoke build system extracts and relocates each individual function without linking them together, and cannot easily use the object file mode as the original compiler emits object files in COFF format rather than ELF. As such, my use case requires diffs to be in binary mode and cannot use map files.

Add support for skipping the map file lookup in `dump_binary()` when a file is specified on the command line.

This commit changes the long form of the `-f` flag from `--objfile` to `--file`. If this change in name is likely to break existing workloads, `--objfile` can either be added as an alias for `--file` or we can just use `--objfile` always.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1498,8 +1498,8 @@ def dump_binary(
     binfile = config.file or project.myimg
     if not project.baseimg or not binfile:
         fail("Missing myimg/baseimg in config.")
-    if config.make and project.myimg is not None:
-        run_make(project.myimg, project)
+    if config.make:
+        run_make(binfile, project)
     if not os.path.isfile(binfile):
         fail(f"Not able to find binary file: {binfile}")
     start_addr = maybe_eval_int(start)
--- a/diff.py
+++ b/diff.py
@@ -1495,12 +1495,10 @@ def dump_objfile(
 def dump_binary(
     start: str, end: Optional[str], config: Config, project: ProjectSettings
 ) -> Tuple[str, ObjdumpCommand, ObjdumpCommand]:
-    if not project.baseimg or not project.myimg and not config.file:
+    binfile = config.file or project.myimg
+    if not project.baseimg or not binfile:
         fail("Missing myimg/baseimg in config.")
-    binfile = project.myimg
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -1495,16 +1495,23 @@ def dump_objfile(
 def dump_binary(
     start: str, end: Optional[str], config: Config, project: ProjectSettings
 ) -> Tuple[str, ObjdumpCommand, ObjdumpCommand]:
-    if not project.baseimg or not project.myimg:
+    if not project.baseimg or not project.myimg and not config.file:
         fail("Missing myimg/baseimg in config.")
+    binfile = project.myimg
+    if config.file is not None:
+        binfile = config.file
```
simonlindholm, line 1502:
```py
    binfile = config.file or project.myimg
    if not project.baseimg or not binfile:
        fail("Missing myimg/baseimg in config.")
```

diff.py
```diff
@@ -1495,12 +1495,10 @@ def dump_objfile(
...
-    if not project.baseimg or not project.myimg and not config.file:
+    binfile = config.file or project.myimg
+    if not project.baseimg or not binfile:
         fail("Missing myimg/baseimg in config.")
-    binfile = project.myimg
-    if config.file is not None:
-        binfile = config.file
-    if config.make:
+    if config.make and project.myimg is not None:
         run_make(project.myimg, project)
```
simonlindholm, line 1502: hm, doesn't it make more sense to run make on `binfile` btw?

asdf2jkl: My thoughts were that the file specified on the command line would be updated/produced as a byproduct of running make on `myimg` as `binfile` likely isn't a make target.

I suppose my thought process was skewed by my specific use case (which doesn't use `-m`) with one build command that outputs a ton of files. It would be more logical to try to run `make binfile` instead.

simonlindholm:
Can you resolve the merge conflict? I'll merge after that

PR #120, opened 2023-08-15
Expand --stop-at-ret to stop at variable return count

asdf2jkl:
When diffing two very different functions in binary mode, the `-s` command-line switch is essential in order to give the diff algorithm a chance to output something useful. If it is not used, the diff algorithm cannot effectively match up the source and target instructions and outputs garbage.

The `-s` switch causes the diff to stop at the first return instruction it finds. This is useful for producing more relevant diffs, but comes at the cost of truncating output when there are several return instructions in the same function. That truncation removes essential context for the user viewing the diff.

This patch modifies the `--stop-at-ret`/`-s` command-line flag to optionally take N, an integer argument, and truncate the diff at the Nth return instruction found. The existing behaviour of the flags is retained by defaulting N to one when there is no integer argument passed along with the flag.

The benefits of this flag are outlined below with an example function (a deliberately different `DerefSym` from AMS 3.10, for those curious)

Here is an example of a poorly-matched function with `-s` not used:
image
Note how the diff algorithm only detects a partial instruction match __two__ functions later in the binary than the target function, making that match and all subsequent matches useless from a comparison point of view.

Here is an example of the same function, but using `-s` this time:
image
The diff has a little bit more context, but now we have a conditional branch on line two with a target outside of the listing. That won't do in most cases of function decompilation.

Lastly, here is the same function with the correct amount of returns passed (`-s 2`):
image
The functions are still extremely different, but asm-differ now gives a high-quality look at how they are different.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -212,9 +212,14 @@ if __name__ == "__main__":
         "-s",
         "--stop-at-ret",
         dest="stop_at_ret",
-        action="store_true",
-        help="""Stop disassembling at the first return instruction.
-        Some functions have multiple return points, so use with care!""",
+        metavar="N",
+        type=int,
+        default=None,
+        const=1,
+        nargs='?',
+        help="""Stop disassembling at the Nth return instruction.
+        Defaults to not stopping on returns, or stopping at the first return
+        encountered if N is not specified.""",
     )
     parser.add_argument(
         "-i",
@@ -449,7 +454,7 @@ class Config:
     show_branches: bool
     show_line_numbers: bool
     show_source: bool
-    stop_at_ret: bool
+    stop_at_ret: int
     ignore_large_imms: bool
     ignore_addr_diffs: bool
     algorithm: str
@@ -2405,6 +2410,7 @@ def process(dump: str, config: Config) -> List[Line]:
     source_lines = []
     source_filename = None
     source_line_num = None
+    rets_remaining = config.stop_at_ret
 
     i = 0
     num_instr = 0
@@ -2586,7 +2592,9 @@ def process(dump: str, config: Config) -> List[Line]:
         source_lines = []
 
         if config.stop_at_ret and processor.is_end_of_function(mnemonic, args):
-            break
+            rets_remaining -= 1
+            if rets_remaining == 0:
+                break
 
     processor.post_process(output)
     return output
--- a/diff.py
+++ b/diff.py
@@ -213,9 +213,8 @@ if __name__ == "__main__":
         "--stop-at-ret",
         dest="stop_at_ret",
         action="count",
-        help="""Stop disassembling when a return instruction is encountered.
-        As some functions have multiple return points, you can pass this
-        multiple times.""",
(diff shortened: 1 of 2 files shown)


simonlindholm:
> The existing behaviour of the flags is retained by defaulting N to one when there is no integer argument passed along with the flag.

It isn't quite: if you do `-so` it will claim `invalid int value: 'o'` instead of parsing it as -s -o, and if you do `./diff.py -wombs func_800ADBB0` it will try to parse `func_800ADBB0` as an int. I think these are fairly common usages.

Possible solution: make it possible to pass -s multiple times, incrementing the counter by 1 each time?

diff.py
```diff
@@ -449,7 +454,7 @@ class Config:
     show_branches: bool
     show_line_numbers: bool
     show_source: bool
-    stop_at_ret: bool
+    stop_at_ret: int
```
simonlindholm, line 457:
```suggestion
    stop_at_ret: Optional[int]
```

diff.py
```diff
@@ -2586,7 +2592,9 @@ def process(dump: str, config: Config) -> List[Line]:
         source_lines = []
 
         if config.stop_at_ret and processor.is_end_of_function(mnemonic, args):
```
simonlindholm, line 2594: (requiring `rets_remaining` or `rets_remaining is not None` here for mypy to be happy)

asdf2jkl:
> > The existing behaviour of the flags is retained by defaulting N to one when there is no integer argument passed along with the flag.
> 
> It isn't quite: if you do `-so` it will claim `invalid int value: 'o'` instead of parsing it as -s -o, and if you do `./diff.py -wombs func_800ADBB0` it will try to parse `func_800ADBB0` as an int. I think these are fairly common usages.
> 
> Possible solution: make it possible to pass -s multiple times, incrementing the counter by 1 each time?

Drats, I didn't test with that.
My original implementation used a counting flag, but I found that it was unintuitive. I'm used to the `-v`/`-vv`/ `-vvv` usecase for counting flags, and the `-s` flag's usage does not match that pattern. But at least the number of function returns should be a small number.

simonlindholm:
Explaining it via example may be best, something like "You can also pass -ss to stop at the second return instruction, -sss at the third, etc."

simonlindholm:
Sure!

test.py
```diff
@@ -24,13 +24,13 @@ def get_config(self) -> diff.Config:
...
-            ignore_addr_diffs=True,
-            algorithm="levenshtein",
-            reg_categories={},
+            show_line_numbers=False, 
+            show_source=False, 
+            stop_at_ret=None, 
+            ignore_large_imms=False, 
+            ignore_addr_diffs=True, 
+            algorithm='levenshtein', 
+            reg_categories={},
```
simonlindholm, line 33:
```suggestion
            show_line_numbers=False,
            show_source=False,
            stop_at_ret=None,
            ignore_large_imms=False,
            ignore_addr_diffs=True,
            algorithm="levenshtein",
            reg_categories={},
```

PR #122, opened 2023-08-19
Introduce flag to bypass watch-mode prompt

asdf2jkl:
While it is recommended not to use -w without -m, there are uses for it.
However, calling asm-differ from another program is complicated by having to work around the confirmation of having to run make manually.

Add a -y flag to automatically skip that prompt.


simonlindholm:
Oh, this is fair. Should probably mention the option in the prompt, or perhaps get rid of the prompt altogether and force you to use -y.

PR #125, opened 2023-09-09
Fix indirect calls like calll *0x123456

roblabla:
This would fail on `calll *0x123456` instructions with an exception claiming it failed to parse `*0x123456` as an int because of the prefixed star.

I'm not sure if this fix is correct - I have absolutely no clue what much of the code in here is trying to do. But it does allow the differ to continue.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2571,7 +2571,7 @@ def process(dump: str, config: Config) -> List[Line]:
         if (
             mnemonic in arch.branch_instructions or is_text_relative_j
         ) and symbol is None:
-            x86_longjmp = re.search(r"\*(.*)\(", args)
+            x86_longjmp = re.search(r"\*([0-9a-fA-Fx]+)", args)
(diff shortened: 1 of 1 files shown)


simonlindholm:
I have little idea also. But it seems reasonable.

Now the code tries first to search for `r"\*(.*)\("`, and then `r"\*(.*)"` as a fallback if it fails, performing the same logic if either matches. But the second should imply the first, so can we just get rid of the first check?

roblabla:
You'd think so, but actually no! I initially did that as well, but it broke on instructions of the form `whatever *0x12(%eax)`, because now we would try to run `int("0x12(%eax)")` which is also wrong ^^.

simonlindholm:
Ah! Maybe we can do `\*([0-9x]+)`?

roblabla:
hmm, yeah that might work

PR #127, opened 2023-09-10
Fix stuff like call *(%eax)

roblabla:
So turns out the fix in #125 broke stuff like `call *(%eax)`. So I'm resubmitting the original fix (along with a pretty comment to explain what any of this is doing because god knows if I ever have to change this again in a couple months, I'll have forgotten how it all works again :joy:)


ethteck:
Is any of this specific to att syntax? I'm hoping nor

roblabla:
It is.  In Intel Syntax, you'd have to look for stuff like `call    dword ptr [edi]`.

PR #129, opened 2023-10-05
Handle ATT x86 register dereference

OmniBlade:
Prevents exception on register dereferences for x86 ASM.


simonlindholm:
Ugh, even more AT&T special-casing... This seems to cause a test failure for some reason.

OmniBlade:
I have no idea what that test failure means or why the change would cause it to fail :/

PR #130, opened 2023-10-05
Show some x86 reloc info correctly.

OmniBlade:
Due to small sample size used to develop this probably isn't exhaustive.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2653,14 +2653,14 @@ def process(dump: str, config: Config) -> List[Line]:
             x86_longjmp = re.search(r"\*(.*)\(", args)
             if x86_longjmp:
                 capture = x86_longjmp.group(1)
-                if capture != "":
+                if capture != "" and capture.isnumeric():
                     branch_target = int(capture, 16)
             else:
                 # Then, we try to match the global deref in a separate regex.
                 x86_longjmp = re.search(r"\*(.*)", args)
                 if x86_longjmp:
                     capture = x86_longjmp.group(1)
-                    if capture != "":
+                    if capture != "" and capture.isnumeric():
                         branch_target = int(capture, 16)
                 else:
                     branch_target = int(args.split(",")[-1], 16)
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1786,12 +1786,24 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
...
+        # Example *0x0(,%edx,4)
+        # Match 0x0 part to replace
+        addr_imm = re.search(r"(^|(?<=\*))0x0", args)
         if not addr_imm:
-            assert False, f"failed to find address immediate for line '{prev}'"
+            # match end of line with a hex string
+            # Example %edi,0
+            # Example call a2f
+            # Example movb $0x0,0x0
+            addr_imm = re.search(r"0?x?[\*0-9a-f]+$", args)
```
simonlindholm, line 1803: 0?x? -> (?:0x)? ?

diff.py
```diff
@@ -1786,12 +1786,24 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
...
+        if "WRTSEG" in row: # ignore WRTSEG (watcom)
+            return prev, None
         repl = row.split()[-1]
         mnemonic, args = prev.split(maxsplit=1)
-
-        addr_imm = re.search(r"(?<!\$)0x[0-9a-f]+", args)
+        # Example 0x0,0x8(%edi)
+        # Example 0x0,%edi
+        # Example *0x0(,%edx,4)
+        # Match 0x0 part to replace
```
simonlindholm, line 1796: Do you get 0x0 even if the reloc has an addend, e.g., when compiling
```
int x[2];
int f() { return x[1]; }
int* g() { return &x[1]; }
```
?

OmniBlade: Okay, I've refactored the reloc handling again to handle those kinds of cases, any other cases you can think of that I might need to handle?

PR #133, opened 2023-12-04
Fix PowerPC GCC branches which contain ".text+"

1superchip:
Removes ".text+0x" from branches
Reverts previous commit which had incorrect logic


simonlindholm (approved):
Sorry for the delay. Looks good.

Issue #121, opened 2023-08-16
Identical numeric values considered different in diff


marijnvdwerf:
<img width="664" alt="260849477-7ee25f4d-7ebb-4471-96db-9b2953721d4a" src="https://github.com/simonlindholm/asm-differ/assets/737603/981843b0-0f7c-46dc-b3bd-4ebe9984eb9c">

(Source: https://decomp.me/scratch/xtJVv)

The target has 0x1, but the current version of the assembly displays 1. I'd expect it to not count as an _actual_ difference.

Originally reported as https://github.com/decompme/decomp.me/issues/816

AngheloAlf:
This is not a bug, this is an actual difference. This is caused because of objdump's pseudo instructions.
You can see the actual difference in the scratch you posted by clicking on "Disable pseudo instructions".

simonlindholm:
That said, it would be nice if we could convert hex li instructions into `addiu x, zero, imm` to make this clearer, it's very confusing right now.

AngheloAlf:
iirc the asm produced by psx compilers needs to be post-processed so it can match. The toolchain is a bit yucky.
It would be cool if asm-differ had an option to treat cases like this as the same instruction, but I don't think this should be enabled by default because this can be a real real difference on other compilers (it has at least happened on IDO)

ethteck:
I think Simon just means we should convert the hex li instructions in the output of this tool, not actually editing the input. but yeah I agree after this convo we shouldn't consider these equal, probably ever really

ethteck:
<img width="702" alt="image" src="https://github.com/simonlindholm/asm-differ/assets/2985314/5b8ba6de-e6e2-416b-9c81-075660389979">
(with pseudos off)

It seems we are treating some as equal, so that feels inconsistent

PR #134, opened 2023-12-17
Ignore hex/decimal differences in 'li' on MIPS

mkst:
Is this the best way to tackle this? I put it behind a guard so we aren't trying to do a regex match for every line...

With this change:
image
Without:
image

**[edit]** Just seen the issue here https://github.com/simonlindholm/asm-differ/issues/121, which is pretty much what this PR is trying to address - outside of pseudo operations, its just trying to ignore 0x6 vs 6...

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1577,9 +1577,9 @@ class AsmProcessorMIPS(AsmProcessor):
 
     def _normalize_arch_specific(self, mnemonic: str, row: str) -> str:
         if mnemonic == "li":
-            # only consider values of 0-9 equivalent
-            regex = re.compile(f"(0x[0-9])")
-            return re.sub(regex, lambda m: str(int(m.group(1), 16)), row)
+            # only consider values of 1-9 equivalent
+            regex = re.compile(f",(0x[1-9])$")
+            return re.sub(regex, lambda m: f",{int(m.group(1), 16)}", row)
         return row
 
     def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
--- a/diff.py
+++ b/diff.py
@@ -470,6 +470,7 @@ class Config:
     penalty_reordering = 60
     penalty_insertion = 100
     penalty_deletion = 100
+    ignore_equivalent_immediates = False
 
 
 def create_project_settings(settings: Dict[str, Any]) -> ProjectSettings:
@@ -1576,9 +1577,8 @@ class AsmProcessorMIPS(AsmProcessor):
         self.seen_jr_ra = False
 
     def _normalize_arch_specific(self, mnemonic: str, row: str) -> str:
-        if mnemonic == "li":
-            # only consider values of 1-9 equivalent
-            regex = re.compile(f",(0x[1-9])$")
+        if self.config.ignore_equivalent_immediates and mnemonic == "li":
+            regex = re.compile(f",(0x[1-9])$")  # only 1 thru 9
             return re.sub(regex, lambda m: f",{int(m.group(1), 16)}", row)
         return row
 
@@ -2185,7 +2185,9 @@ MIPS_SETTINGS = ArchSettings(
     proc=AsmProcessorMIPS,
 )
 
-MIPSEL_SETTINGS = replace(MIPS_SETTINGS, name="mipsel", big_endian=False)
+MIPSEL_SETTINGS = replace(
+    MIPS_SETTINGS, name="mipsel", big_endian=False, arch_flags=["-m", "mips:3000"]
+)
 
 MIPSEE_SETTINGS = replace(
     MIPSEL_SETTINGS, name="mipsee", arch_flags=["-m", "mips:5900"]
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1575,6 +1575,13 @@ def __init__(self, config: Config) -> None:
         super().__init__(config)
         self.seen_jr_ra = False
 
+    def _normalize_arch_specific(self, mnemonic: str, row: str) -> str:
+        if mnemonic == "li":
+            return re.sub(
+                f"(-?0x[0-9a-fA-F]+)", lambda m: str(int(m.group(1), 16)), row
+            )
```
AngheloAlf, line 1582: Using a compiled regex with `re.compile` (https://docs.python.org/3/library/re.html#re.compile) would be better

AngheloAlf (commented):
Could you change this so this feature can be turned on and off, since this can be a very real diff on N64 (at least on IDO).

See the convo at #134

mkst:
I've been playing around with this and only 1 thru 9 appear as 1 thru 9 (rather than 0x1 thru 0x9). Negative values are decimal, and values above 10 are hexidecimal. Does this still need to go behind an option - I know that `-1` and `0xFF` are different values under IDO, but this PR won't treat them as equivalent (after my latest changes)

AngheloAlf:
Here's a scratch where the difference between the two matters
https://decomp.me/scratch/xv8L5

This doesn't match when linked, so asm-diff showing no diffs would be misleading.

AngheloAlf:
btw, for some reason when turning off pseudo instructions asm-differ shows `dli $t1, 0x1` instead of `ori $t1, $zero, 0x1`. This seems to be some weird bug on asm-differ, since this doesn't happen on objdump

mkst:
Wow, I had no idea IDO was this awful 😅 I'll stick it behind an option, one for tomorrow

simonlindholm:
Have you looked into patching compiler/assembler to avoid this diff instead? Would be nice to avoid option bloat. (Though I'm not really opposed to adding one, either.)

simonlindholm:
> btw, for some reason when turning off pseudo instructions asm-differ shows dli $t1, 0x1 instead of ori $t1, $zero, 0x1. This seems to be some weird bug on asm-differ, since this doesn't happen on objdump

Guessing this is due to objdump flags. I don't know which ones decomp.me passes

mkst:
Could be a mips1 vs mips2 thing. The assembler is gnu `as` here... I'll see whether it makes any difference to change the value in maspsx before it goes to as... But that won't help the N64 SN stuff.

mkst:
Setting the arch for mipsel turns the dli into ori:
```py
MIPSEL_SETTINGS = replace(MIPS_SETTINGS, name="mipsel", big_endian=False, arch_flags=["-m", "mips:3000"])
```
image

mkst:
Very open to opinions on the name of the config option, `ignore_equivalent_immediates` was the best my tired brain could come up with.

AngheloAlf:
> > btw, for some reason when turning off pseudo instructions asm-differ shows dli $t1, 0x1 instead of ori $t1, $zero, 0x1. This seems to be some weird bug on asm-differ, since this doesn't happen on objdump
> 
> Guessing this is due to objdump flags. I don't know which ones decomp.me passes

It doesn't seem to be a decomp.me issue since it happens locally too

diff.py
```diff
@@ -1575,6 +1576,12 @@ def __init__(self, config: Config) -> None:
         super().__init__(config)
         self.seen_jr_ra = False
 
+    def _normalize_arch_specific(self, mnemonic: str, row: str) -> str:
+        if self.config.ignore_equivalent_immediates and mnemonic == "li":
+            regex = re.compile(f",0x([1-9])$")  # only 1 thru 9
```
AngheloAlf, line 1581: The idea of compiling a regex is to compile it only once, not every time the function is called

mkst: I was googling and read:

> **Is it worth using Python’s re.compile()?**
 As you know, Python always internally compiles and caches regexes whenever you use them anyway (including calls to search() or match()), so using compile() method, you’re only changing when the regex gets compiled.

But sure, I'll stick it outside the function

mkst:
I'll need to add this to `create_config()` to be able to use it in decomp.me...

mkst:
OK chaps, please re-review now that I've made some changes; again, if you can think of a better (shorter) name for this option please shout!

mkst:
This PR will become less important for anyone using `maspsx` because I've just pushed a [commit](https://github.com/mkst/maspsx/commit/34bd6f784c37bb0befd6d9d67178a5cdc33814bc) to expand the `li` instruction.

simonlindholm:
You didn't add a cmdline flag, so this doesn't work as is. Actually, I think it should probably be a project setting, since it feels assembler-/project-specific.

Given the commit to maspsx though, is there still a use-case for this? Or should we close it and reopen if the need comes up again?

mkst:
I'm happy to just close this one, glædelige jul 🎅

PR #135, opened 2023-12-19
Add 'black' to ci

mkst:


simonlindholm:
Seems you renamed the wrong job, looks okay otherwise. Two small doubts:
- should we pin the black version? Unclear either way, probably won't ever actually matter
- the readme says it's fine to skip black, in the hopes of making the barrier to contributions lower. Is that a policy that should be kept? Maybe keep it but still run black in CI?

mkst:
I've now fixed the job name and pinned black. I don't feel it's a big ask for contributors to run `black` but guess it's fine for the test to fail in CI and contributions still be merged?

simonlindholm:
Can you update the black version in https://github.com/simonlindholm/asm-differ/blob/main/.pre-commit-config.yaml to the same thing?

.github/workflows/unit-tests.yml
```diff
@@ -7,7 +7,7 @@ on:
 permissions: read-all
 
 jobs:
-  tests:
+  black:
```
simonlindholm, line 10:
```suggestion
  tests:
```

PR #139, opened 2024-02-07
Speed up tab completion: process map file as bytes + mmap

Dragorn421:
The title and code comments should explain everything I hope

To be clear this breaks tabcompleting in the esoteric case of a map file using a (very-far-from-ascii) not-self-synchronizing-encoding, which sounds like it shouldn't happen in practice. Cf note in the readme

The speed up for me is from ~17ms processing to 6-8ms, timing with:

<details>

<summary>patch for time</summary>

```diff
diff --git a/diff.py b/diff.py
index 0763091..705e08d 100755
--- a/diff.py
+++ b/diff.py
@@ -67,6 +67,7 @@ if __name__ == "__main__":
         def complete_symbol(
             prefix: str, parsed_args: argparse.Namespace, **kwargs: object
         ) -> List[str]:
+            import time
             import locale
 
             if not prefix or prefix.startswith("-"):
@@ -79,6 +80,7 @@ if __name__ == "__main__":
             if not mapfile:
                 return []
             completes = []
+            start = time.time()
             encoding = locale.getencoding()
             # We assume the encoding is self-synchronizing,
             # meaning that for example finding bytes corresponding to the space
@@ -114,6 +116,7 @@ if __name__ == "__main__":
                         match = data[pos:endPos]
                         pos = data.find(search, endPos)
                     completes.append(match.decode(encoding))
+            argcomplete.warn((time.time() - start) * 1000, "ms")
             return completes
 
         setattr(start_argument, "completer", complete_symbol)
```

</details>

Relevant python docs:
https://docs.python.org/3/library/locale.html#locale.getencoding
https://docs.python.org/3/library/mmap.html

Note that
`mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ)`
written like that, is portable (across Windows / Unix-y)

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -87,22 +87,29 @@ if __name__ == "__main__":
             if not mapfile:
                 return []
             completes = []
-            with open(
-                mapfile,
-                encoding=MAPFILE_ENCODING,
-                errors=MAPFILE_ENCODING_ERROR_HANDLER,
-            ) as f:
-                data = f.read()
+            encoding = MAPFILE_ENCODING
+            # We assume the encoding is self-synchronizing,
+            # meaning that for example finding bytes corresponding to the space
+            # character, is equivalent to finding a space character.
+            # This is true for ASCII and UTF-8 data, for example.
+            # This allows processing the map file as bytes instead of entirely
+            # processing it as decoded text, which is slow.
+            space = " ".encode(encoding)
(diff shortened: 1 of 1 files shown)


Dragorn421:
To be rebased onto main after #140 is merged

PR #140, opened 2024-02-07
Mapfile encoding: use UTF-8 with error handler "replace"

Dragorn421:
Tiny can of worms opened by #139

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -29,6 +29,16 @@ def static_assert_unreachable(x: NoReturn) -> NoReturn:
     raise Exception("Unreachable! " + repr(x))
 
 
+# The encoding to use for reading the map file to text.
+# It is expected the map file will be ASCII-only or mostly,
+# which UTF-8 is compatible with.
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -71,13 +81,17 @@ def complete_symbol(
                 # skip reading the map file, which would
                 # result in a lot of useless completions
                 return []
-            config: Dict[str, Any] = {}
-            diff_settings.apply(config, parsed_args)  # type: ignore
-            mapfile = config.get("mapfile")
-            if not mapfile:
+            settings: Dict[str, Any] = {}
+            diff_settings.apply(settings, parsed_args)  # type: ignore
+            project = create_project_settings(settings)
```
simonlindholm, line 86: Does this work? I think create_project_settings is defined too far below.

simonlindholm (approved):

diff.py
Dragorn421: yeah I choked

PR #141, opened 2024-02-10
Add mips:4000 arch (psp)

mkst:
Any opinions on the naming convention?


simonlindholm:
> Any opinions on the naming convention?

Looks a bit odd but I don't have any better ideas and it doesn't matter much. Let's just roll with it.

PR #145, opened 2024-02-17
Add `bal` to `MIPS_BRANCH_INSTRUCTIONS`

AngheloAlf:
Without this is not possible to use diff.py on functions with `bal` instructions because it raises an exception while parsing objdump's output.

This is caused because `reloc_addend_from_imm` tries to convert the hex without prefix to int with `int(imm, 0)`.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2037,6 +2037,7 @@ MIPS_BRANCH_INSTRUCTIONS = MIPS_BRANCH_LIKELY_INSTRUCTIONS.union(
         "bltz",
         "bc1t",
         "bc1f",
+        "bal",
     }
 )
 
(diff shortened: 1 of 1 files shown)


simonlindholm:
It's possible it would be better to treat it the same as `jal` to avoid branch arrows, but this works. Out of curiosity, where have you found bal's? I haven't come across compilers that emit them.

AngheloAlf:
By treating them the same way as `jal` you mean adding `bal` here?
https://github.com/simonlindholm/asm-differ/blob/734edde4cfa42eb45b971734efc0c4ed6e724da0/diff.py#L2259-L2260

I'm trying to match handwritten asm that uses `bal`s. I haven't seen a compiler that emits `bal`s either

PR #148, opened 2024-02-25
i686: regex: Support using the FS segment register for immediates

### LINKED ISSUE #143: i386 differ fails when diffing against code with old C++ header contents in the context?
See this scratch: https://decomp.me/scratch/RbmNx
(Note: the header in question is a preprocessed context of bstring.h from the Hewlett Packard C++ STL.)

``Diff error: Error running asm-differ: failed to find address immediate for line 'add	%al,(%eax)'``

ChrisNonyminus:
This should help fix part of #143.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1885,6 +1885,10 @@ class AsmProcessorI686(AsmProcessor):
             offset = True
 
         if not addr_imm:
+            addr_imm = re.search(r"(^|(?<=\*)|(?<=\%fs\:))0x[0-9a-f]+", args)
+            offset = True
+            
+        if not addr_imm:
             assert False, f"failed to find address immediate for line '{prev}'"
 
         start, end = addr_imm.span()
--- a/diff.py
+++ b/diff.py
@@ -1885,9 +1885,9 @@ class AsmProcessorI686(AsmProcessor):
             offset = True
 
         if not addr_imm:
-            addr_imm = re.search(r"(^|(?<=\*)|(?<=\%fs\:))0x[0-9a-f]+", args)
+            addr_imm = re.search(r"(^|(?<=\*)|(?<=\%[fgdecs]s\:))0x[0-9a-f]+", args)
             offset = True
-            
+
         if not addr_imm:
             assert False, f"failed to find address immediate for line '{prev}'"
 


simonlindholm:
Do you have some example objdump output that this matches?

ChrisNonyminus:
> Do you have some example objdump output that this matches?

Not sure what you mean matches, but here's an example in a local instance of decomp.me with this applied that the diff works on:
Cheese_Sun-25Feb24_08 11

ChrisNonyminus:
Here's a raw objdump with mingw's objdump if you're curious:
```

Disassembly of section .text:

Disassembly of section .text:

00000000 <??0cRZCmdLine@@QAE@PBD@Z>:
??0cRZCmdLine@@QAE@PBD@Z():
   0:   push   %ebp
   1:   mov    %esp,%ebp
   3:   push   $0xffffffff
   5:   push   $0x0
                        6: dir32        __ehhandler$??0cRZCmdLine@@QAE@PBD@Z-0x2a
   a:   mov    %fs:0x0,%eax
                        c: dir32        __except_list
...
```
Note that with msdosdjgpp, the current objdump executable for win9x/win32 on decomp.me, it gives this error:
```

i386-pc-msdosdjgpp-objdump: backend/local_files/current.o: unrecognized storage class 105 for *UND* symbol `??_EcRZCmdLine@@UAEPAXI@Z'
i386-pc-msdosdjgpp-objdump: failed to read symbol table from: backend/local_files/current.o
i386-pc-msdosdjgpp-objdump: error message was: file format not recognized
```
And I should clarify what exactly this PR fixes and how, by the way:
When getting around that aforementioned msdosdjgpp error by making a certain virtual destructor pure or non-virtual, msdosdjgpp's objdump works, but with the current asm-differ on decomp.me, it gives this:
```
code.c
Diff error: Error running asm-differ: failed to find address immediate for line 'mov	%fs:0x0,%eax'
```
Basically, the regex currently does not account for uses of the fs segment register for immediates, and this PR fixes specifically that.
@simonlindholm

simonlindholm:
Should we cover other segment registers as well, and not just fs?

ChrisNonyminus:
Hmm, maybe.

ChrisNonyminus:
However, I haven't yet run into any uses of other segment registers.

ChrisNonyminus:
Okay, here you go.

Issue #56, opened 2021-09-17
Add an API for getting the JSON diff as an object instead of a string


zbanks:
Motivated by decomp.me.

In #52 I added the `JsonFormatter` for returning the diff results as a JSON blob. This is fine for CLI use, but decomp.me imports `diff.py` directly and uses it as a library. If we're worried about diff performance in decomp.me, `diff.py` could expose a function for getting the diff result as a `dict`, rather than as a serialized `str`.

PR #151, opened 2024-03-10
Add PythonFormatter that JsonFormatter inherits from

mkst:
Closes https://github.com/simonlindholm/asm-differ/issues/56.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -916,6 +916,10 @@ class PythonFormatter(Formatter):
         return NotImplemented
 
     def table(self, data: TableData) -> str:
+        # This method is unused by this formatter
+        return NotImplemented
+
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -973,6 +973,13 @@ def serialize(text: Optional[Text]) -> List[Dict[str, Any]]:
                     output_row[column_name] = column
             output_rows.append(output_row)
         output["rows"] = output_rows
+        return output
```
simonlindholm, line 976: The return type annotation of `table()` becomes wrong now. Can you rename the method to something else, leaving a `table` method that just raises a NotImplementedError?

(mypy ought to catch this error, I believe)

simonlindholm (changes requested):

PR #152, opened 2024-03-29
Add function symbols to diff output

1superchip:
This adds function symbols to the output while diffing. The function symbols will be highlighted when different and contribute to the score.
Added an option of "-d" or "--diff-function-symbols" to enable function symbols in the diff.

Different Labels
Multiple Functions

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1126,7 +1126,7 @@ def restrict_to_function(dump: str, fn_name: str) -> str:
     try:
         # Find the start of the line that contains "<fn_name>:"
         ind = dump.rfind("\n", 0, dump.index(f"<{fn_name}>:"))
-        return dump[ind:]
+        return dump[ind + 1 :]
     except ValueError:
         return ""
 
--- a/diff.py
+++ b/diff.py
@@ -1125,8 +1125,8 @@ def run_make_capture_output(
 def restrict_to_function(dump: str, fn_name: str) -> str:
     try:
         # Find the start of the line that contains "<fn_name>:"
-        ind = dump.rfind("\n", 0, dump.index(f"<{fn_name}>:"))
-        return dump[ind + 1 :]
+        ind = dump.rfind("\n", 0, dump.index(f"<{fn_name}>:")) + 1
+        return dump[ind:]
     except ValueError:
         return ""
 
@@ -2601,11 +2601,11 @@ def process(dump: str, config: Config) -> List[Line]:
                 
                 output.append(
                     Line(
-                        mnemonic=function_name, # this is needed for score changes
-                        diff_row=function_name, # this is needed to be set for highlighting
-                        original=function_name, # this is the value that is printed
+                        mnemonic="<label>",
+                        diff_row=function_name,
+                        original=function_name,
                         normalized_original=function_name,
-                        scorable_line=function_name,
+                        scorable_line="label " + function_name,
                     )
                 )
             continue
--- a/diff.py
+++ b/diff.py
@@ -2598,7 +2598,13 @@ def process(dump: str, config: Config) -> List[Line]:
         row = lines[i]
         i += 1
 
-        if config.diff_function_symbols:
+        if not row:
+            continue
+
+        if config.diff_function_symbols == True:
+            # If diffing function symbols is enabled
+            # Check if the line contains a function label
+            # If so, create a line to diff for the label
 
             # Regex the current row to check for "OFFSET <SYM>:"
             funcNameMatch = re.match(r"[0-9A-Fa-f]+ <(.+)>:", row)
@@ -2615,10 +2621,9 @@ def process(dump: str, config: Config) -> List[Line]:
                         scorable_line="...",
                     )
                 )
-
-        if not row:
-            continue
-
+                continue
+        
+        # Filter out function label lines "OFFSET <SYM>:"
         if re.match(r"^[0-9a-f]+ <.*>:$", row):
             continue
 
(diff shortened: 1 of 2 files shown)


diff.py
```diff
@@ -1114,9 +1123,22 @@ def run_make_capture_output(
 
 
 def restrict_to_function(dump: str, fn_name: str) -> str:
+
+    # Regex objdump's output to find the beginning of the line which contains the offset and symbol
+    # "OFFSET <SYM>:"
+    restrictFunctionRegex = f"[0-9A-Fa-f]+ <{fn_name}>:"
```
simonlindholm, line 1129: fn_name needs escaping to be part of a regex, and also searching for a regex seems a bit unnecessarily slow. Can we keep searching as before but then search backwards for a newline?

diff.py
```diff
@@ -2579,6 +2601,29 @@ def process(dump: str, config: Config) -> List[Line]:
         if not row:
             continue
 
+        if config.diff_function_symbols == True:
```
simonlindholm, line 2604:
```suggestion
        if config.diff_function_symbols:
```

diff.py
```diff
@@ -2579,6 +2601,29 @@ def process(dump: str, config: Config) -> List[Line]:
...
+                        diff_row=funcSym, # this is needed to be set for highlighting
+                        original=funcSym, # this is the value that is printed
+                        normalized_original="...",
+                        scorable_line="...",
+                    )
+                )
+                continue
+        
+        # Filter out function label lines "OFFSET <SYM>:"
         if re.match(r"^[0-9a-f]+ <.*>:$", row):
```
simonlindholm, line 2627: This is the same check, right? Can you change the logic so it's
```
if a:
  if b:
    ...
  continue
```
instead of
```
if b:
  if a:
    ...
    continue
if a:
  continue
```
?

diff.py
```diff
@@ -2579,6 +2601,29 @@ def process(dump: str, config: Config) -> List[Line]:
...
+            # If diffing function symbols is enabled
+            # Check if the line contains a function label
+            # If so, create a line to diff for the label
+
+            # Regex the current row to check for "OFFSET <SYM>:"
+            funcNameMatch = re.match(r"[0-9A-Fa-f]+ <(.+)>:", row)
+            
+            if funcNameMatch:
+                # Add the function symbol to the diff output
+                funcSym = funcNameMatch.groups()[0] + ":"
```
simonlindholm, line 2614: snake_case

diff.py
```diff
@@ -2579,6 +2601,29 @@ def process(dump: str, config: Config) -> List[Line]:
...
+                output.append(
+                    Line(
+                        mnemonic=funcSym, # this is needed for score changes
+                        diff_row=funcSym, # this is needed to be set for highlighting
+                        original=funcSym, # this is the value that is printed
+                        normalized_original="...",
+                        scorable_line="...",
```
simonlindholm, line 2621: `...` doesn't seem like a good value, can they use funcSym too? Or is the intent behind not using it for scorable_line to avoid mismatches impact scores?

diff.py
```diff
@@ -577,6 +585,7 @@ def create_config(args: argparse.Namespace, project: ProjectSettings) -> Config:
         ignore_addr_diffs=args.ignore_addr_diffs,
         algorithm=args.algorithm,
         reg_categories=project.reg_categories,
+        diff_function_symbols=args.diff_function_symbols
```
simonlindholm, line 588: (nitpick)
```suggestion
        diff_function_symbols=args.diff_function_symbols,
```

diff.py
```diff
@@ -2578,8 +2588,26 @@ def process(dump: str, config: Config) -> List[Line]:
...
+            function_name = function_label_match.groups()[0] + ":"
 
-        if re.match(r"^[0-9a-f]+ <.*>:$", row):
+            if config.diff_function_symbols:
+                # If diffing function symbols is enabled
+                # Add the symbol to the diff output
+                
+                output.append(
+                    Line(
+                        mnemonic=function_name, # this is needed for score changes
```
simonlindholm, line 2604: This is the value that gets used for lining things up. In a case like
```
func_abc:
ins1
ins2
ins3
```
vs
```
func_xyz:
ins2
ins3
```
we want `func_abc:` and `func_xyz:` to line up, instead of `func_abc:` appearing on its own and `func_xyz` lining up with `ins1`, which would be an equally good scoring diff outcome if mnemonics don't match. So I would suggest putting a constant `mnemonic="<label>"` to counteract this.

`scorable_line` might need to be `"label " + function_name` or something like that for scoring to work in that case.

diff.py
```diff
@@ -2578,8 +2588,26 @@ def process(dump: str, config: Config) -> List[Line]:
...
+                        diff_row=function_name, # this is needed to be set for highlighting
+                        original=function_name, # this is the value that is printed
```
simonlindholm, line 2606: This is the wrong place for those comments. Ones in the `Line` class would be useful though, for sure, although they would require quite some care to get right so don't feel obliged

diff.py
```diff
@@ -1115,7 +1124,8 @@ def run_make_capture_output(
 
 def restrict_to_function(dump: str, fn_name: str) -> str:
     try:
-        ind = dump.index("\n", dump.index(f"<{fn_name}>:"))
+        # Find the start of the line that contains "<fn_name>:"
+        ind = dump.rfind("\n", 0, dump.index(f"<{fn_name}>:"))
         return dump[ind + 1 :]
```
simonlindholm, line 1129:
```suggestion
        ind = dump.rfind("\n", 0, dump.index(f"<{fn_name}>:")) + 1
        return dump[ind:]
```

Pretty neat that this works even if the line is the very first one. It's not often -1 sentinels work in your favor like that.

PR #163, opened 2024-06-02
Annotate and add branch targets for ARM32 jump tables

abaresk:
ARM32-Thumb encodes jump tables as `.short {offset}`, where `offset` gets added to the program counter to perform the jump. Unfortunately, this encoding obfuscates the control flow of the jump table. Furthermore, `objdump` frequently interprets the entries as ARM32 instructions, or groups 2 jump entries together into a single `.word` entry.

This change adds architecture-specific preprocessing, primarily to split `.word` entry lines into 2 `.short` lines. In preprocessing, it also converts entries that were interpreted as ARM32 instructions back to `.short`, and it includes the target address for each jump entry in comments.

In post-processing, this change populates `Line.branch_target` for each jump entry.

Before: https://pastebin.com/raw/TxXKDcwM (address `0x44`)
After: https://pastebin.com/raw/mBmRE4Zt

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1211,7 +1211,8 @@ def preprocess_objdump_out(
             + out
         )
 
-    return out
+    processor = config.arch.proc(config)
+    return processor.preprocess_objdump(out)
 
 
 def search_build_objects(objname: str, project: ProjectSettings) -> Optional[str]:
@@ -1602,14 +1603,15 @@ def dump_binary(
     )
 
 
-# Example: "ldr r4, [pc, #56]    ; (4c <AddCoins+0x4c>)"
-ARM32_LOAD_POOL_PATTERN = r"(ldr\s+r([0-9]|1[0-3]),\s+\[pc,.*;\s*)(\([a-fA-F0-9]+.*\))"
-
-
 # The base class is a no-op.
 class AsmProcessor:
     def __init__(self, config: Config) -> None:
         self.config = config
+
+    # Called during run_objdump() for arch-specific normalization. Runs before
+    # diff-processing, i.e. process().
+    def preprocess_objdump(self, objdump: str) -> str:
+        return objdump
 
     def pre_process(
         self, mnemonic: str, args: str, next_row: Optional[str]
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1780,8 +1782,100 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+ARM32_COMPARE_IMM_PATTERN = r"cmp\s+(r[0-9]|1[0-3]),\s+#(\w+)"
+
+# Example: "add pc, r1"
+ARM32_JUMP_TABLE_START = r"add\s+pc,\s*r"
+
+# Examples:
+#  - "44:   00060032    .word   0x00060032"
+#  - "48:   0032        .short  0x0032"
+#  - "4a:   0032        movs    r2, r6"
+#  - "9e:   00be        lsls    r6, r7, #2"
```
simonlindholm, line 1795: The fact that we're adding a dependency on the bytes being included in the output here is a bit sad, I was hoping to be able to add --no-show-raw-insn for all architectures. Oh well. An alternative would be to do real ELF parsing for this instead, like in `parse_elf_rodata_references`, but I guess it's more fragile.

diff.py
```diff
@@ -2491,6 +2597,17 @@ class ArchSettings:
     M68K_SETTINGS,
 ]
 
+def immediate_to_int(immediate: str) -> int:
```
simonlindholm, line 2600: I believe this can be replaced by `int(immediate.lstrip("#"), 0)`

simonlindholm (approved):
Nice! The string munging feels like it adds some fragility and could break in the future from unrelated changes, so it would be really nice if we had some regression testing... unfortunately we don't, but maybe you could at least dump the original objdump output or (better) the .o files here in the PR for when it's hopefully added in the future? (Well, we do have test.py, but IMO it's a bit too complicated to add tests for.)

abaresk:
Sure thing, here are the object files that are referenced in the PR description: 
[arm32-jumptable.zip](https://github.com/user-attachments/files/15755961/arm32-jumptable.zip)

Issue #161, opened 2024-05-26
Relocations aren't used in some function calls


1superchip:
Some function calls don't contain the name of their target but rather the offset of the target. This stems from the assembler not emitting relocations to static functions.

The left side is an object file produced by the assembler while the right side is produced by the compiler.

Current diff:
![image](https://github.com/simonlindholm/asm-differ/assets/57328807/c6d9ee2e-ed30-40e7-86d1-f5753d2a954e)

Target assembly:
![image](https://github.com/simonlindholm/asm-differ/assets/57328807/32fa2020-cd33-49e1-9e87-57fdf8056c8e)
Note that `AsyncResourceClearState` and `AsyncResourceSetState` are local.

objdump output:
```
00000724 <AsyncResourceClearState>:
     724:       38 00 00 00     li      r0,0
     728:       b0 00 00 00     sth     r0,0(0)
                        728: R_PPC_EMB_SDA21    gAsyncResourceStatus
     72c:       4e 80 00 20     blr

00000730 <AsyncResourceSetState>:
     730:       a0 00 00 00     lhz     r0,0(0)
                        730: R_PPC_EMB_SDA21    gAsyncResourceStatus
     734:       54 00 05 1e     rlwinm  r0,r0,0,20,15
     738:       b0 00 00 00     sth     r0,0(0)
                        738: R_PPC_EMB_SDA21    gAsyncResourceStatus
     73c:       a0 00 00 00     lhz     r0,0(0)
                        73c: R_PPC_EMB_SDA21    gAsyncResourceStatus
     740:       7c 00 1b 78     or      r0,r0,r3
     744:       b0 00 00 00     sth     r0,0(0)
                        744: R_PPC_EMB_SDA21    gAsyncResourceStatus
     748:       4e 80 00 20     blr

0000074c <AsyncResourceStateBusy>:
     74c:       7c 08 02 a6     mflr    r0
     750:       90 01 00 04     stw     r0,4(r1)
     754:       94 21 ff f0     stwu    r1,-16(r1)
     758:       98 61 00 08     stb     r3,8(r1)
     75c:       b0 81 00 0a     sth     r4,10(r1)
     760:       4b ff ff c5     bl      724 <AsyncResourceClearState>
     764:       38 60 10 00     li      r3,4096
     768:       4b ff ff c9     bl      730 <AsyncResourceSetState>
     76c:       a0 60 00 00     lhz     r3,0(0)
```

[...]

PR #165, opened 2024-06-10
Use comments to apply relocations to certain instructions

1superchip:
Addresses https://github.com/simonlindholm/asm-differ/issues/161

This is an attempt to fix an issue with the assembler not emitting relocations to static functions.

This may cause issues when absolute values are used such as in this scratch: https://decomp.me/scratch/KBprJ
The input assembly of the scratch is `bl 4` which objdump disassembles to image

How asm-differ outputs this with my changes: 
image

Other instructions could be added but would cause issues such as conditional/unconditional branches to labels within functions
This scratch has labels which the assembler emits: https://decomp.me/scratch/hKpwa

objdump of the input assembly from the scratch: 
image
These branch instructions have a comment which matches the comment regex that is used in AsmProcessorPPC.pre_process so the offsets (which we want) would get replaced with the label if these instructions are included.

This is why AsmProcessorPPC.pre_process checks for `bl`

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1613,7 +1613,7 @@ class AsmProcessor:
         return objdump
 
     def pre_process(
-        self, mnemonic: str, args: str, next_row: Optional[str]
+        self, mnemonic: str, args: str, next_row: Optional[str], comment: str|None
     ) -> Tuple[str, str]:
         return mnemonic, args
 
@@ -1694,7 +1694,7 @@ class AsmProcessorMIPS(AsmProcessor):
 
 class AsmProcessorPPC(AsmProcessor):
     def pre_process(
-        self, mnemonic: str, args: str, next_row: Optional[str]
+        self, mnemonic: str, args: str, next_row: Optional[str], comment: str|None
     ) -> Tuple[str, str]:
         if next_row and "R_PPC_EMB_SDA21" in next_row:
             # With sda21 relocs, the linker transforms `r0` into `r2`/`r13`, and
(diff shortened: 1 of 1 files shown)


diff.py
```diff
@@ -1731,7 +1731,12 @@ def pre_process(
             splitArgs[-1] = next_row.split(".text+0x")[-1]
             args = ",".join(splitArgs)
 
-        if comment is not None and mnemonic == "bl":
+        if (
+            comment is not None
+            and next_row is not None
+            and re.search(self.config.arch.re_reloc, next_row) is None
```
simonlindholm, line 1737: Isn't the logical thing `next_row is None or re.search(self.config.arch.re_reloc, next_row) is None`? Not that it matters in practice of course, but...

1superchip: next_row will always be set with either a relocation or the next instruction if it isn't None.

```
        next_line = lines[i] if i < len(lines) else None
        mnemonic, args = processor.pre_process(mnemonic, args, next_line, comment)
```

I check if it isn't None to validate it. If it isn't None, then it's checked with the reloc regex to determine if it has a relocation or not.

If it doesn't have a relocation in it, then continue with the mnemonic check

simonlindholm: Yes, I get that. But you're checking for the absence of a relocation, in which case `next_row is None` should be accepted, not rejected

1superchip: Ah, yes. You are correct that the code should continue with the next checks if `next_row` is `None` as it doesn't contain a relocation.

diff.py
```diff
@@ -1731,6 +1731,19 @@ def pre_process(
...
+            comment is not None
+            and (
+                next_row is None
+                or re.search(self.config.arch.re_reloc, next_row) is None
+            )
+            and mnemonic == "bl"
+        ):
+            # if the mnemonic is bl and the comment doesn't match
+            # <.text+0x...> replace the args with the contents of the comment
+            if re.search(r"<.+\+0x[0-9a-fA-F]+>", comment) == None:
```
simonlindholm, line 1744:
```suggestion
            if re.search(r"<.+\+0x[0-9a-fA-F]+>", comment) is None:
```
more pythonic

simonlindholm (approved):

PR #166, opened 2024-06-19
Fix common ARM32 assembly discrepancies

abaresk:
This change normalizes the assembly output to remove 2 common sources of diff scoring penalties:

1. Extra padding appearing at the end of function definitions (line 62 in before)
2. Optional destination register omitted by assembler (line 28 in before)

Before: https://pastebin.com/raw/CybxVDe2
After: https://pastebin.com/raw/BMXyfD5s

[arm32-discrepancies.zip](https://github.com/user-attachments/files/15905941/arm32-discrepancies.zip)


diff.py
```diff
@@ -1895,11 +1914,23 @@ def _jump_table_entries_count(self, raw_lines: List[str], line_no: int) -> int:
...
         return 0
 
+    def pre_process(
+        self, mnemonic: str, args: str, next_row: Optional[str], comment: Optional[str]
+    ) -> Tuple[str, str]:
+        arg_parts = args.split()
+        # Normalize instructions that omit the destination register.
+        if len(arg_parts) == 2 and any(
+            [insn in mnemonic for insn in ARM32_DEST_OPTIONAL]
+        ):
```
simonlindholm, line 1929:
```suggestion
        if len(arg_parts) == 2 and any(ins in mnemonic for ins in ARM32_DEST_OPTIONAL):
```
making use of generators.

A bit silly to split on whitespace instead of `", "`, but works fine in this case.

simonlindholm (approved):

simonlindholm:
> Optional destination register omitted by assembler (line 28 in before)

Surprising to hear of this sources of mismatches. Is it different in the binary, e.g. some arm vs thumb thing? Or what causes objdump to either one or the other?

abaresk:
The issue is with the assembler used in decomp.me. Objdump is correct in disambiguating these 2 cases, as the instructions have different encodings:

```
1c40: adds r0, r0, #1
3001: adds r0, #1
```

Unfortunately the assembler used by decomp.me `arm-none-eabi-as` coerces instances of the former case into the latter.

abaresk:
That makes sense, I'll report back after investigating the effort on adding proprietary assemblers to decomp.me.

I did some spelunking into modern binutils to see if there's any flags to disable this behavior. Unfortunately, there did not seem to be any -- from what I can tell, this is [baked into](https://github.com/facebookarchive/binutils/blob/a535268b59862077d95f34f1572ac0bce0b428c7/gas/config/tc-arm.c#L23160-L23166) to the relocation-processing stage of the assembler (`md_apply_fix()`).

PR #178, opened 2025-01-27
Add x86 architecture

Gillou68310:


diff.py
```diff
@@ -2817,7 +2851,7 @@ def process(dump: str, config: Config) -> List[Line]:
         line_num = eval_line_num(line_num_str.strip())
 
         # TODO: use --no-show-raw-insn for all arches
-        if arch.name == "i686" or arch.name == "aarch64":
+        if arch.name == "i686" or arch.name == "x86" or arch.name == "aarch64":
```
simonlindholm, line 2854:
```suggestion
        if "--no-show-raw-insn" in arch.arch_flags:
```

diff.py
```diff
@@ -2018,40 +2018,58 @@ def _normalize_adrp_differences(self, mnemonic: str, row: str) -> str:
...
+        # Example movb $0x0,0x4(%si)
+        if not addr_imm:
+            addr_imm = re.search(r"(?<=,)0x[0-9a-f]+", args)
+            offset = True
+
         # Example 0x4,%eax
         # Example $0x4,%eax
         if not addr_imm:
-            addr_imm = re.search(r"(^|(?<=\*)|(?<=\$))0x[0-9a-f]+", args)
+            addr_imm = re.search(r"(^|(?<=\*)|(?:\$))0x[0-9a-f]+", args)
```
simonlindholm, line 2072: this will cause $'s to start appearing in the output, right? is that expected?

diff.py
```diff
@@ -2538,14 +2565,20 @@ class ArchSettings:
     re_large_imm=re.compile(r"-?[1-9][0-9]{2,}|-?0x[0-9a-f]{3,}"),
     re_sprel=re.compile(r"-?(0x[0-9a-f]+|[0-9]+)(?=\((%ebp|%esi)\))"),
     re_imm=re.compile(r"-?(0x[0-9a-f]+|[0-9]+)|([\?$_][^ \t,]+)"),
-    re_reloc=re.compile(r"R_386_|dir32|DISP32|WRTSEG|OFF32|OFFPC32"),
+    re_reloc=re.compile(
+        r"R_386_|dir32|DISP32|WRTSEG|OFF32|OFFPC32|OFF16|OFFPC16|:\s+3\s"
```
simonlindholm, line 2569: huh, what's up with the `:\s+3\s`?

simonlindholm (approved):

diff.py
Gillou68310:
````  1a:   push   %ax
  1b:   push   %ds
  1c:   mov    $0x0,%ax
                        1d: OFF16       _tpal
                        1d: WRTSEG      DGROUP
  1f:   push   %ax
  20:   lcall  $0x0,$0x0
                        21: 3   _write_pels
  25:   add    $0x8,%sp```

Comment (Gillou68310) [reply diff.py:2569]: lcall somehow emit relocs with '3' instead of a named reloc type

File: diff.py (hunk 2 above)
Comment (Gillou68310) [reply diff.py:2072]: Nope it's actually to remove the $ from the output

Comment (simonlindholm) [reply diff.py:2072]: Ah, I see, because we look at span.start()/.end() to know what part of the instruction to remove! I was looking at
```py
        if offset:
            repl = f"{repl}+{addr_imm.group()}"
```
which I guess still carries the `$`, but maybe that's fine. Could do an lstrip() otherwise.

simonlindholm (approved):

simonlindholm:
(will merge once you push the `if "--no-show-raw-insn" in arch.arch_flags:` change)

Gillou68310:
Done, I also added the 'SEG' reloc type (Using Huge memory model)
Thanks for reviewing btw!

diff.py
```diff
@@ -2304,8 +2332,9 @@ class ArchSettings:
     "bns-",
 }
 
-I686_BRANCH_INSTRUCTIONS = {
+X86_BRANCH_INSTRUCTIONS = {
     "call",
+    "lcall",
```
Gillou68310, line 2337: Ok this was a bad idea, it tries to resolve branch target for it and fails...

PR #183, opened 2025-02-10
[WIP]Rework

Gillou68310:


Gillou68310:
@1superchip

simonlindholm:
@1superchip I gave you repo access, you can squash&merge this when it looks good to you

1superchip:
We need to add `call` to `instructions_with_address_immediates` on https://github.com/simonlindholm/asm-differ/blob/fc49910aae85150c768f6c44928a5efe6e34daca/diff.py#L2612

diff.py
```diff
@@ -2080,14 +2080,14 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
-            addr_imm = re.search(r"(?<=,)0x0+(?=\(.*\))", args)
+            addr_imm = re.search(r"(?<=,)(?:0x)?0+(?=\(.*\))", args)
 
         # Example 0x0,0x8(%edi)
         # Example 0x0,%edi
         # Example *0x0(,%edx,4)
         # Example $0x0,0x4(%edi)
         if not addr_imm:
-            addr_imm = re.search(r"(^\$?|(?<=\*))0x0", args)
+            addr_imm = re.search(r"(^\$?|(?<=\*))(?:0x)?0", args)
```
1superchip, line 2090: This regex change results in a difference when matching the following args: `$0x400,%ecx`.
The old regex wouldn't match anything in this whereas the new regex matches `$0`.

1superchip: The regex is finding `$0` in `$0x400`. 

We want a regex that finds `0x0` OR `0`. The regex `(^\$?|(?<=\*))(?:0x)?0` find `0` in `0x3` which we don't want. 
Find `0x0` or (`0` but not `0` in `0x...`)
this regex seems to do what is wanted:
`(^\$?|(?<=\*))(?:0x)?0(?!x)`

<https://regex101.com/r/9011a8/1> `(^\$?|(?<=\*))(?:0x)?0(?!x)`
<https://regex101.com/r/zHCgMf/1> `(^\$?|(?<=\*))(?:0x)?0`

PR #192, opened 2025-03-20
Prevent crashes on line-parsing errors

### LINKED ISSUE #179: Make asm-differ more resilient to parsing errors
I'm not sure how feasible it even is, but it'd be cool to have asm-differ bail on individual lines it can't handle and just fall back to some un-parsed representation, rather than totally error out

From @simonlindholm:
> yes, making asm-differ more resilient would be fantastic
> make it try-except the processing of every line, and do something smart in the except handler that effectively just emits a single error line

example failure: https://decomp.me/scratch/XlGeq

Requested originally by @MrSapps

ethteck:
Fixes #179 

There's probably a bit more work to be done here, but this is a start

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -3032,7 +3032,18 @@ def process(dump: str, config: Config) -> List[Line]:
                         branch_target = int(args.split(",")[-1], 16)
         except Exception as e:
             # If we fail to parse the line, at least emit something rather than crashing.
-            original = str(e)
+            output.append(
+                Line(
+                    mnemonic="ERROR",
+                    diff_row=row,
+                    original=str(e),
+                    normalized_original=str(e),
+                    scorable_line=str(e),
+                )
+            )
+            num_instr += 1
+            source_lines = []
+            continue
 
         output.append(
             Line(
--- a/diff.py
+++ b/diff.py
@@ -1364,7 +1364,7 @@ def parse_elf_rodata_references(
     data: bytes, config: Config
 ) -> List[Tuple[int, int, str]]:
     e_ident = data[:16]
-    if e_ident[:4] != b"\x7FELF":
+    if e_ident[:4] != b"\x7fELF":
         return []
 
     SHT_SYMTAB = 2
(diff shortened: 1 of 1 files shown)


ethteck:
I recommend viewing this PR with [?w=1 appended to the URL](https://github.com/simonlindholm/asm-differ/pull/192/files?w=1)

1superchip:
Does this print anything if an error does occur?

ethteck:
image
Example with errors

diff.py
```diff
@@ -2847,184 +2847,192 @@ def process(dump: str, config: Config) -> List[Line]:
...
+                    x86_longjmp = re.search(r"\*(.*)", args)
+                    if x86_longjmp:
+                        capture = x86_longjmp.group(1)
+                        if capture != "" and capture.isnumeric():
+                            branch_target = int(capture, 16)
+                    else:
+                        branch_target = int(args.split(",")[-1], 16)
+        except Exception as e:
+            # If we fail to parse the line, at least emit something rather than crashing.
+            original = str(e)
```
simonlindholm, line 3035: Hm, we're constructing a Line with a whole bunch of properties after this, but they aren't set if an exception is thrown. I guess we should assign empty values to them before the start of the try block?

simonlindholm: or better, do `output.append(Line(...)); continue` like we do in some other cases within the try, and just ignore all of the properties normally passed when constructing the Line

ethteck: Sure - I also added the num_instr +=1 and the source_lines = [] that also happens in the normal case

diff.py
```diff
@@ -2847,184 +2847,203 @@ def process(dump: str, config: Config) -> List[Line]:
...
+                        branch_target = int(args.split(",")[-1], 16)
+        except Exception as e:
+            # If we fail to parse the line, at least emit something rather than crashing.
+            output.append(
+                Line(
+                    mnemonic="ERROR",
+                    diff_row=row,
+                    original=str(e),
+                    normalized_original=str(e),
+                    scorable_line=str(e),
```
simonlindholm, line 3041: If you clear source_lines, it would make sense to add source_lines here so its data is not lost

PR #195, opened 2025-04-30
Handle x86 MSVC jumptables

1superchip:
This is a draft pr that implements handling of MSVC jumptables for x86. It currently requires diff function symbols to work properly as that is needed for the labels to be processed but it should not require that in the long run. 

A small issue with this is that if there are multiple text sections, arrows will be shown in each text section. There was also an issue with objdump not emitting proper information for jumptables in a 2nd text section if the first section had a jump table.

objdump output of the jumptable:
image

objdump output of the assembly that is referenced from the jumptable:
image

An image of the final result:
image

We need <$Lnum> in the 2nd image to get the location.


diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+
+                # Remove "-0x..." from older objdump relocations
+                jump_table_target = re.sub(r"-0x[0-9a-f]+", "", orig_jump_table_target)
+
+                jump_table_targets.append(jump_table_target)
+
+                # If a line begins with ".dword", it is a jumptable
+                was_previous_jumptable_entry = True
+
+                target_line_num: Union[int, None]
```
simonlindholm, line 2237:
```suggestion
                target_line_num: Optional[int]
```

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+                try:
+                    target_line_num = lines[
+                        labels[jump_table_target + ":"] + 1
+                    ].line_num
+                except:
+                    target_line_num = None
+
+                if target_line_num:
```
simonlindholm, line 2245:
```suggestion
                if target_line_num is not None:
```

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+        labels: Dict[str, int] = {}  # line number for each label
+
+        was_previous_jumptable_entry: bool = False
+
+        for index, line in enumerate(lines[:]):
+            if line.original.startswith(".dword"):
+
+                orig_jump_table_target: str = line.original.split(".dword")[1].strip()
+
+                # Remove "-0x..." from older objdump relocations
```
simonlindholm, line 2229: What does older mean, are we talking about different compiler or objdump versions, or is it older in the sense of us not yet having processed them?

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+                    target_line_num = lines[
+                        labels[jump_table_target + ":"] + 1
+                    ].line_num
+                except:
+                    target_line_num = None
+
+                if target_line_num:
+                    line.original = line.original.replace(
+                        orig_jump_table_target, hex(target_line_num).replace("0x", "")
+                    )
```
simonlindholm, line 2248: Can we set `line.original = ".dword\t" + hex(...)` instead? Search-replace always feels fragile to me

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+
+                    line.branch_target = target_line_num
+            elif was_previous_jumptable_entry == True:
```
simonlindholm, line 2251:
```suggestion
            elif was_previous_jumptable_entry:
```

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+        # and adds branch arrows to the jump table entries
+
+        # This will have errors for multiple text sections as objdump (at least 2.38) doesn't emit
+        # the correct labels for sections that are not the first
+
+        jump_table_targets: List[str] = []  # target_label
+
+        labels: Dict[str, int] = {}  # line number for each label
+
+        was_previous_jumptable_entry: bool = False
```
simonlindholm, line 2222:
```suggestion
        was_previous_jumptable_entry = False
```

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+        # Handle jump tables produced by MSVC.
+        # MSVC jump table labels are in the format of "$L{NUM}"
+
+        # Removes extra "add\t%al,(%eax)" which correspond to an existing jumptable entry
+        # and adds branch arrows to the jump table entries
+
+        # This will have errors for multiple text sections as objdump (at least 2.38) doesn't emit
+        # the correct labels for sections that are not the first
+
+        jump_table_targets: List[str] = []  # target_label
```
simonlindholm, line 2218: this seems to be unused

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+                    line.original = line.original.replace(
+                        orig_jump_table_target, hex(target_line_num).replace("0x", "")
+                    )
+
+                    line.branch_target = target_line_num
+            elif was_previous_jumptable_entry == True:
+                # if the previous line was a jumptable and the current line is "add\t%al,(%eax)"
+                # this is the 2nd half of the previous jumptable entry
+                if line.original == "add\t%al,(%eax)":
+                    lines.remove(line)
```
simonlindholm, line 2255: `.remove(...)` loops over the entire `lines` list and checks dataclass equality, which is slow and can have false positives. We could instead e.g. do
```py
indices_to_delete = []
for index, ... in enumerate(lines):
    if ...:
        indices_to_delete.append(index)
for index in reversed(indices_to_delete):
    del lines[index]
```

diff.py
```diff
@@ -2201,6 +2204,65 @@ def process_reloc(self, row: str, prev: str) -> Tuple[str, Optional[str]]:
...
+
+                was_previous_jumptable_entry = False
+            elif line.mnemonic == "<label>" and line.original.startswith("$L"):
+                # If this is a label that is an MSVC jumptable label "$L..."
+                # add it to the labels list
+                labels[line.original] = lines.index(line)
```
simonlindholm, line 2261: ... and then this would let us use `index` instead of finding the line anew because `lines` has not yet been re-indexed

diff.py
1superchip: Looks like it. I have the targets stored to it but they are never read from it.

diff.py
1superchip: Older objdump versions such as 2.39 emit `-0x...` in relocations relative to the section start
image
image

The binutils that decomp.me uses is `GNU objdump (GNU Binutils) 2.41.90.20240122` which does not emit the `-0x...`.

diff.py
1superchip: Resolved in faa3f549d14e10a8e0c9ed75a8131e0730cac3b0

diff.py
simonlindholm: Seems like a good idea to me to handle them. I'd extend the comment with a mention of objdump 2.39 to make it more understandable

PR #199, opened 2025-09-10
Add row styling to PythonFormatter

mkst:
I would like to make it easy to color add/remove/changed rows in decomp.me, e.g.

image

I don't particularly like my approach hence creating this as a draft to get some input. I feel that it *should* be straightforward to know if a row is an add/remove/change (ignoring any specific register/label diffs) - without having to iterate over the cells...

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -948,6 +948,13 @@ class PythonFormatter(Formatter):
                 return []
             return [serialize_format(s, f) for s, f in text.segments]
 
+        def get_row_style(text: Text) -> Optional[str]:
+            for (_, f) in text.segments:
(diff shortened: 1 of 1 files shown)


simonlindholm:
The ability to get JSON for three-way diffs makes this classification a bit weird; you can e.g. have something that is a deletion in `curr` but a change in `prev`. I don't see the problem with looping over the cells to determine this information? Especially if this JSON gets sent over the wire for decomp.me it makes sense to not bloat it.

simonlindholm:
I think I'd prefer for this to be on the decomp.me side, yes. I think decomp.me should already have as good a data source as asm-differ does at this point, so I don't see the advantage of putting logic in asm-differ. Unless I'm missing something? I don't understand the DIFF_REMOVED detection problem enough, should just be a matter of detecting "<" I would have thought.

PR #200, opened 2025-11-06
SH improvements

Mc-muffin:
This was supposed to be "a small contribution copying from ARM" but due to SH objects not having a way to signal objump what's data and what's code (like ARM does) it ended up exploding in size a bit, I'm not particularly good writing big chunks of code so I'm not sure how mergeable/readable/reviewable this is, I apologize in advance.

This is also my first real contribution to asm-differ so I'm sure there's stuff I missed or didn't get.

Among the changes:
- Added jump table detection, as generated by SHC and the scratch in #113 
- Added literal pool detection, literal pool moves get an inline comment like ARM does
- Related to this, literal pool mis-disassembly gets corrected to `.word`/`.long`
- Some relocation support, only the ones I noticed in some Dreamcast scratches, though SHC only seems to generate `R_SH_DIR32` relocs

While setting up the PR I noticed issue #113 which would be addressed by this, though this PR solves it by highly massaging what objdump puts out instead of the other idea of using a different disassembler's output.

Some before and afters:
<details>

| before | after |
 |--------|------|
| image | image |
| image |  image |
| image | image |

</details>

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2395,7 +2395,7 @@ class AsmProcessorSH2(AsmProcessor):
                 else:
                     mov_value = -1
 
-                    # Can be None if direct values are used but the pool 
+                    # Can be None if direct values are used but the pool
                     # doesn't get included in the asm
                     if mov_match.group(5) is not None:
                         mov_value = int(mov_match.group(5), 16)


diff.py
```diff
@@ -2202,13 +2202,374 @@ def is_end_of_function(self, mnemonic: str, args: str) -> bool:
...
+        self._relocs: Dict[int, str] = {}
+
+    def preprocess_objdump(self, objdump: str) -> str:
+        new_lines = []
+        lines = objdump.splitlines()
+
+        # Detect endianess, as there's no special marker for what is data
+        # and what is code (like in ARM) getting the endian is needed to
+        # correct literal pool bytes read as instructions
+        is_be = self._test_endian(lines)
```
simonlindholm: If you wanted to simplify this, we actually have the raw ELF available for parsing as well, from where the endian should be possible to read out from the header. Or (better?) you could do like "sh4" vs "sh4el", where `arch.big_endian` (accessible from `config.arch` in `__init__`) is statically known based on what the user passes in.

diff.py
```diff
@@ -2202,13 +2202,374 @@ def is_end_of_function(self, mnemonic: str, args: str) -> bool:
...
+                        jtbl_addr = int(jtbl_match.group(3), 16)
+                        break
+
+                if is_mova and addr not in self._relocs:
+                    # Remove from imm table if present
+                    if jtbl_addr in self._imms:
+                        del self._imms[jtbl_addr]
+
+                    # Search up to 10 lines before
+                    end = 10 if i >= 10 else i
```
simonlindholm:
```suggestion
                    end = min(i, 10)
```

PR #211, opened 2026-03-03
Add xenon arch settings for xbox 360

1superchip:
Adds the settings for powerpc-xenon for xbox 360.

maybe `-m` and `-M` should be explicit?

Let me know about the names.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -2987,6 +2987,10 @@ PPC_SETTINGS = ArchSettings(
     proc=AsmProcessorPPC,
 )
 
+PPC_XENON_SETTINGS = replace(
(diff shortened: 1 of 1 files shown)


simonlindholm:
Is the thing here that this arch is expected to be used with a custom objdump that sets the -M default to something that decodes instructions in a different way than broadway?

I don't have any better ideas for naming, this works for me.

1superchip:
The current settings specify `-m broadway` for `ppc` which causes the custom powerpc-xenon-pe-objdump to not disassemble certain instructions. By removing that, it disassembles the instructions. I would like to have the switch explicit but I'm not sure the exact values. 

Maybe I should add a comment that it is meant for a custom objdump?

